PR #11, opened 2026-06-30
Remove local trial limits (macOS + Windows) — local is free & unlimited

ray-amjad:
## Summary

HyperWhisper is fully open source, so the **local, on-device** transcription path no longer has a trial gate. Local recording and model downloads are now **unconditionally free and unlimited** on every platform. HyperWhisper **Cloud** (server-side paid transcription) is unchanged — it remains the paid moat, enforced server-side.

## What changed

**Rust core (authoritative):**
- `hw-license/usage.rs`: `check_limits` returns unlimited for every status; `can_download_model` always `true`. `Limits` / `DEFAULT_*` / `cache.rs` kept **inert** to preserve a stable FFI surface. Unit + golden tests rewritten to assert unlimited.
- Rebuilt the checked-in macOS static lib; regenerated bindings are **byte-identical** (FFI unchanged). The Windows `hyperwhisper_core.dll` rebuilds from cargo in CI — no checked-in binary.

**macOS:**
- Removed the recording daily-limit gate, both `recordTranscriptionTime` calls, and the model-download paywall/alert.
- Deleted `LicenseUsageTracker.swift` + `ConfigService.swift`; slimmed `LicenseManager` (removed trial delegation + `loadRemoteConfig`).
- Reframed the sidebar "Upgrade to Pro" CTA into a **HyperWhisper Cloud** credits CTA.
- Updated all 40 `Localizable.strings` (removed trial keys, added Cloud-CTA keys).

**Windows (mirror):**
- Removed both recording gates + toasts, both `RecordTranscriptionTime` calls, the model-download guard, and model-download counting.
- Deleted `TrialUsageCard`, `ModelsTrialWarning`; reframed `TrialSidebarActions` → Cloud credits CTA.
- Deleted `LicenseUsageTracker.cs` + `ConfigService.cs`; removed `LoadRemoteConfigAsync` and dead trial delegation from `LicenseManager.cs`.
- Updated all 40 `Strings*.resx` (removed trial keys, added Cloud-CTA key).

**Docs / site:**
- Rewrote `licensing.mdx` around free/unlimited local + Cloud credits; fixed stale trial copy in `troubleshooting.mdx` and the license email; marked `/api/config` (`trial_daily_limit_seconds`) deprecated/unused.
- No `shared-backup` schema change (no Mode/vocabulary fields touched).

## Verification

- ✅ **Rust:** `cargo test -p hw-license` green (41 tests); `build-bindings.sh` → empty diff on `.swift` / `.cs` bindings (FFI unchanged).
- ✅ **macOS:** `xcodebuild` compiles clean (signing-disabled); bundled `Localizable.strings` verified (new keys present, removed keys gone).
- ⚠️ **Windows:** not compile-verified locally (WPF/.NET targets Windows). Grep-verified every removed symbol has zero remaining `.cs`/`.xaml` references; all 40 resx parse as well-formed XML. **The `windows-release` CI job is the authoritative compile check.**
- ✅ **Cloud identity unaffected:** `getTranscriptionIdentifier` (license_key vs device_id) untouched on both platforms.

## Manual checks still needed
- macOS runtime (needs Xcode Cmd+R with a real signing team): record past 5 min, download a 4th+ model, confirm no trial UI, Cloud panel/credits/activation still work.
- Windows: build via CI/PowerShell, then the same manual checks.

## Note for reviewer
`licensing.mdx`'s Cloud-credit *pricing* narrative (150 free / 5,000 bundled) was left untouched since it was recently/deliberately written — worth a glance to confirm it reads right alongside the "local is free" reframe.


ray-amjad:
## Codex P2 fixed — `/api/config` no longer gates legacy clients

Fixed in 303bec1.

**The issue:** `/api/config` still returned `trial_daily_limit_seconds: 300` and `trial_model_download_limit: 3`. We removed the desktop `ConfigService` in this PR, but legacy / un-upgraded clients in the field still fetch this endpoint and apply those numbers as **live trial caps** — so they kept enforcing the old 5-min/day + 3-model gate, contradicting "local is free & unlimited."

**The fix:** the endpoint is intentionally kept for back-compat (per `nextjs/CLAUDE.md`), so rather than delete it we return effectively-unlimited values: both fields now return `2_000_000_000`.

**Why `2_000_000_000` and not `i64::MAX` or `0`:**
- Legacy Windows `ConfigService` parsed these as 32-bit `int` → `i64::MAX` overflows and breaks `System.Text.Json` deserialization. `2_000_000_000` fits in int32 (max 2,147,483,647) with headroom; macOS parsed as 64-bit and is fine either way.
- The Rust core enforces via `used >= limit` and `remaining = (limit - used).max(0)`, so `0`/negative would **block everything** (`0 >= 0` is true). A large positive value is effectively unlimited (~63 years of seconds / 2B models) with no underflow.

New desktop builds don't fetch `/api/config` at all (ConfigService removed); this is purely to neutralize the gate for old clients. Note: legacy clients may hold a cached `300/3` response for up to 6h (`Cache-Control: max-age=21600`) — nothing retroactive we can do, new fetches get the unlimited values.

---

The 5 findings from our `/code-review high` workflow were already fixed in ec5a887 (unused `licenseManager` EnvironmentObject, dead `licenseStatusDescription`, stale Windows header-doc lines, the troubleshooting accordion, and the verifier script asserting deleted `CanStartRecording()`).

@codex review

PR #16, opened 2026-07-01
feat(macos): redesigned 8-step first-run onboarding

ray-amjad:
Redesigned first-run onboarding for the macOS app, built as a delta on the original SwiftUI flow.

**Flow (8 steps):** Welcome → Permissions → Choose Source → Configure → Set Up → **Microphone (new)** → Give it a try → Done.

## What changed
- **Access-key rename** — user-facing "license key" → "access key" in the cloud copy (keys and internal identifiers unchanged), synced across all 40 locales.
- **Test-key buttons on Configure** — "Test access key" (`LicenseManager.validateLicense`, no activation) and "Test API key" (`CloudProviderHealthManager.ensureHealthy`), each with inline results that reset on step re-entry / provider change.
- **Curated on-device model list** — spans both engines (Parakeet V2 *recommended* + V3 + Whisper Base + Large v3 Turbo) with speed/accuracy gauges. New `OnboardingModelSelection` type persists the pick to `Mode.model` and routes downloads to the correct manager (`ParakeetModelManager` vs `WhisperModelManager`).
- **New Microphone step** — device picker + live level meter (new idle-metering plumbing on `AudioRecordingManager`) + "Open Sound Settings" link. The metering preview starts on appear and is released on disappear, on recording start, and on onboarding completion.
- **Give-it-a-try** — the `.onboarding` trigger suppresses auto-paste and surfaces the transcript inline; the global paste setting is untouched.
- **Trimmed Done step** — hero + title + Done only.
- **Nav bookkeeping** — `totalSteps` 7→8; switch/nav/dots/gating re-indexed for the inserted step.

## Verification
- `xcodebuild … CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED** (the default signed build only fails at the signing step — the local team is deliberately kept out of git).
- All 40 `.strings` files lint clean (`plutil -lint`); no residual "license" wording in the renamed cloud fields.
- End-to-end runtime walk is manual via Xcode Cmd+R (CLI can't launch — signing + CloudKit team).


app/macos/hyperwhisper/hyperwhisperApp.swift
```diff
@@ -575,12 +575,22 @@ struct MenuBarIconView: View {
...
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-                appState.showOnboarding = true
+            if PersistenceController.shared.didSeedDefaultModesOnLaunch {
+                // Small delay to ensure the main window is ready before showing sheet
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                    appState.showOnboarding = true
+                }
+            } else {
+                // Existing user (modes already present): treat onboarding as done.
+                hasCompletedOnboarding = true
```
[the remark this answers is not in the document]
ray-amjad: Fixed in ca26635. The root cause was that `didSeedDefaultModesOnLaunch` is only true on the launch that actually seeds — on the next launch an interrupted first run looked like a pre-existing user and got marked complete. Added a durable `@AppStorage("onboardingPending")` flag that captures the fresh-install signal and persists it until onboarding completes, so an interrupted first run is re-shown; genuine existing users (never seeded here) are still marked complete exactly once. Also cleared on completion.

app/macos/hyperwhisper/Views/OnboardingView.swift
```diff
@@ -828,160 +465,215 @@ struct OnboardingView: View {
...
+                // "Test" in this session returned valid. A non-empty key text is no
+                // longer enough — it could be wrong.
+                return licenseManager.licenseStatus == .active || keyValidated
+            case .yourProvider:
+                // BYOK must pass the inline "Test" (health probe returned healthy)
+                // before continuing. A returning user re-presses Test once this
+                // session — an accepted tradeoff for never advancing on a dead key.
+                return keyValidated
+            }
+        case 4: // Set up — source must be genuinely usable (mandatory gate)
```
ray-amjad, line 550: Onboarding is a hard modal with no way out: the sheet is presented with `.interactiveDismissDisabled()` (MainAppView.swift:200), there is no Skip/Later/Quit control anywhere in `navigationButtons`, and this step-4 gate requires the chosen source to be *genuinely usable*.

Concrete scenario — fresh install with no network (plane, captive portal, corporate proxy blocking the API hosts):
- `.onDevice` → model download fails, `isSelectedSourceUsable` stays false
- `.hyperwhisperCloud` → step 3 already blocked (`validateLicense` cannot reach the server, so `keyValidated` never becomes true)
- `.yourProvider` → step 3 already blocked (`ensureHealthy` returns `.unreachable`)

The user is stuck in a non-dismissable sheet and the app is completely unusable until the network comes back. Same shape at step 1 if microphone permission is denied at the OS level and the user does not want to grant it.

Previously `hasCompletedOnboarding` defaulted to `true` so this path was unreachable; flipping the default in this PR makes it reachable for every new install. Please add a "Set this up later" / skip affordance (which can just call `completeOnboarding()` without applying a source), or drop the mandatory gate on step 4.

app/macos/hyperwhisper/hyperwhisperApp.swift
```diff
@@ -575,12 +581,28 @@ struct MenuBarIconView: View {
...
+                onboardingPending = true
+            }
+            if onboardingPending {
+                // Small delay to ensure the main window is ready before showing sheet
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                    appState.showOnboarding = true
+                }
+            } else {
+                // Existing user (modes already present): treat onboarding as done.
+                hasCompletedOnboarding = true
```
ray-amjad, line 599: Ordering bug: `launchMinimized && hasCompletedOnboarding` is evaluated ~20 lines *above* this block (hyperwhisperApp.swift:572), but `hasCompletedOnboarding = true` for the existing-user case is only set here, afterwards.

Scenario: an existing user who has "Launch minimized" enabled updates to this build. On that first launch `hasCompletedOnboarding` still reads its (newly flipped) `false` default, so the minimize branch is skipped and the main window pops open unexpectedly. It self-corrects on the next launch, but every existing menu-bar-only user gets one surprise window.

Fix: resolve the onboarding decision before the `launchMinimized` check, or gate the window-hiding on `!onboardingPending` instead.

app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift
```diff
@@ -0,0 +1,1040 @@
...
+                Image(systemName: isSelected ? "largecircle.fill.circle" : "circle")
+                    .foregroundColor(isSelected ? .accentColor : .secondary)
+
+                VStack(alignment: .leading, spacing: 3) {
+                    HStack(spacing: 8) {
+                        Text(model.displayName)
+                            .font(.system(size: 13.5, weight: .semibold))
+                            .foregroundColor(.primary)
+                        pill(model.kind == .parakeet
+                                ? "onboarding.model.pill.parakeet"
```
ray-amjad, line 399: These three keys exist only in `Base.lproj` — they are missing from all 40 localized `.strings` files (verified: `comm` of the Base vs `fr.lproj` onboarding keys yields exactly `onboarding.model.pill.parakeet`, `onboarding.model.pill.whisper`, `onboarding.model.pill.recommended`).

`String.localized` is `NSLocalizedString(self, comment: "")` with no `value:` default (Extensions/LocalizedString.swift:16), and macOS does **not** fall back to Base.lproj per-key when the user's `.lproj` is present. So in French/German/Japanese/… the pill renders `ONBOARDING.MODEL.PILL.RECOMMENDED` (uppercased by `pill(_:tint:)`) on every row of the on-device model list — the first screen most non-English users see after picking On-Device.

Either add the three keys to every locale or hardcode "Parakeet"/"Whisper" (brand names) and only localize `recommended`.

app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift
```diff
@@ -0,0 +1,1040 @@
...
+        }
+    }
+
+    /// Persist the key (health checks read it from Keychain) then run the shared
+    /// provider health probe — the same path `ProviderKeySheet.testConnection`
+    /// uses. Saving here is harmless: step 4 will simply show "saved".
+    private func testAPIKey() {
+        let key = apiKeyInput.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !key.isEmpty else { return }
+        settingsManager.apiKeys.setAPIKey(apiKeyInput, for: selectedProvider)
```
ray-amjad: `testAPIKey()` writes the key into the Keychain *before* validating it, and never rolls it back on failure.

Scenario: user picks "Your API Key" → OpenAI → pastes a typo'd key → Test → result is `.unauthorized`. The bad key is now permanently in the Keychain for OpenAI. They switch the picker to Deepgram, enter a good key, and finish onboarding. From then on the app holds a dead OpenAI key that will silently fail the first time they select OpenAI in Settings — with a green "configured" state in the provider list.

It also weakens the step-4 gate: `isSelectedSourceUsable(.yourProvider)` is just `hasAPIKey(for: selectedProvider)`, which this write satisfies regardless of the probe result. The only thing keeping an unauthorized key from passing step 4 is the separate `keyValidated` gate on step 3.

Suggest clearing the key (`setAPIKey("", for:)` / delete) when `health != .healthy`.

app/macos/hyperwhisper/Views/OnboardingView.swift
```diff
@@ -543,249 +345,92 @@ struct OnboardingView: View {
...
-                        audioManager.toggleRecordingWithTranscription(trigger: .onboarding)
-                    }) {
-                        VStack(spacing: 8) {
-                            Image(systemName: "mic.circle.fill")
-                                .font(.system(size: 64))
-                            Text(localized: "onboarding.test.start.cta")
-                                .font(.headline)
-                        }
+
+                    Text(appState.lastTranscription)
```
ray-amjad: `appState.lastTranscription` is not only the transcript — the recording flow also stuffs raw English error strings into it:

- `RecordingTranscriptionFlow+StopRecording.swift:289` → `"Error: Audio file could not be read"`
- `RecordingTranscriptionFlow+ErrorHandling.swift:38` → `"Error: Audio file missing for retry"`
- `RecordingTranscriptionFlow+ErrorHandling.swift:76` / `:378` → `"Error: \(error.localizedDescription)"`

Scenario: on the "Give it a try" step the transcription fails (no credits, expired key, provider 500, unreadable audio). The panel condition `!appState.lastTranscription.isEmpty` becomes true and the user sees `Error: The request timed out` rendered under the "You said" heading, with the green ✓ "Shown here only — not pasted into other apps" footer — presented as if it were their transcript. This is the very first transcription a new user ever sees.

Suggest filtering the error sentinel (or reading a dedicated success-only value) before showing the panel, and rendering failures in a distinct error style.

app/macos/hyperwhisper/Views/OnboardingView.swift
```diff
@@ -130,17 +119,23 @@ struct OnboardingView: View {
...
+        .frame(width: 760, height: 580)
         .background(VisualEffectBackground())
         .onAppear {
             setupInitialState()
         }
+        // Re-check microphone permission when the user returns from System Settings
+        // (e.g. after enabling it there) so the mandatory permissions gate unblocks
+        // without them having to re-click. Mirrors HomeView's accessibility re-check.
+        .onReceive(NotificationCenter.default.publisher(for: NSApplication.didBecomeActiveNotification)) { _ in
+            hasMicrophonePermission = AVCaptureDevice.authorizationStatus(for: .audio) == .authorized
```
ray-amjad, line 136: This re-check only refreshes the microphone status — `hasAccessibilityPermission` is left stale.

Scenario: the user taps "Open" on the accessibility row, grants it in System Settings, but the poll started by `pollForAccessibilityPermission()` has already timed out / they granted it before ever pressing the button (common — the app asks for accessibility on an earlier launch). Returning to the app, the mic row flips to ✓ but the accessibility row still shows "Open", so onboarding tells the user the permission is missing when it is not. Unlike the mic path there is no re-poll on activation, so the only recovery is pressing "Open" again.

Cheap fix: also refresh accessibility here.

```suggestion
        .onReceive(NotificationCenter.default.publisher(for: NSApplication.didBecomeActiveNotification)) { _ in
            hasMicrophonePermission = AVCaptureDevice.authorizationStatus(for: .audio) == .authorized
            hasAccessibilityPermission = AccessibilityHelper.shared.hasAccessibilityPermission()
        }
```

app/macos/hyperwhisper/Views/MainAppView.swift
```diff
@@ -196,7 +198,9 @@ struct MainAppView: View {
                 .environmentObject(transcriptionPipeline)
                 .environmentObject(settingsManager)
                 .environmentObject(whisperModelManager)
+                .environmentObject(parakeetModelManager)
```
ray-amjad, line 201: These two new `@EnvironmentObject` requirements are not reflected in the `#Preview` at the bottom of this file (MainAppView.swift:929-935), which still injects only `AppState`/`AudioRecordingManager`/`TranscriptionPipeline`/`SettingsManager`. Any `@EnvironmentObject` that is not supplied traps at runtime, so the MainAppView preview crashes on render. Worth adding `ParakeetModelManager()` and `CloudProviderHealthManager()` (plus the pre-existing `WhisperModelManager`/`LicenseManager` gaps) to the preview.

ray-amjad:

**Result: 4 confirmed, 1 partly true.**

---

### `app/macos/hyperwhisper/Views/OnboardingView.swift:531` — 🟡 **PARTLY TRUE**

**Original claim:** Onboarding is a hard modal (interactiveDismissDisabled, no Skip/Later/Quit) and the step-4 gate requires a genuinely usable transcription source, so a fresh install with no network is permanently stuck with no way out of the app.

**What the code actually does:** MainAppView.swift:204 applies `.interactiveDismissDisabled()` to the `.sheet(isPresented: $appState.showOnboarding)`. OnboardingView.swift:434-467 (`navigationButtons`) renders only Back + Continue/Done — there is no Skip/Later/Quit control, and no `onboarding.skip`/`onboarding.later` key exists in Base.lproj. OnboardingView.swift:531-533 gates step 4 on `isSelectedSourceUsable(source)`, which at :540-555 requires (a) on-device: `whisperModelManager.getModelPath(for:)` != nil — implemented at WhisperModelManager.swift:464-466 as a lookup into `downloadedModels` (disk-downloaded only; no bundled Whisper/Parakeet model), (b) cloud: `licenseManager.licenseStatus == .active`, (c) BYOK: a stored key that step 3 (:529) already requires to have passed a live health probe. All three need network on a fresh install, so an offline first run genuinely cannot advance past step 3/4, and AppState.swift:197-199 suppresses TextDeliveryGate for the whole sheet lifetime so recording is useless meanwhile. `onboardingPending` (hyperwhisperApp.swift:596-602) keeps re-showing it on relaunch.

**Corrected statement:** Accurate up to severity/wording: the user is blocked from *completing onboarding* while offline, not locked inside the process. The onboarding sheet is window-modal (a SwiftUI `.sheet`), not app-modal, and nothing in the app removes or disables the standard Quit menu item, so Cmd+Q still terminates. The block also clears as soon as network is available — it is a hard offline first-run blocker with no skip affordance, not a permanent unrecoverable lock.


### `app/macos/hyperwhisper/hyperwhisperApp.swift:605` — ✅ **CONFIRMED**

**Original claim:** Ordering bug: `launchMinimized && hasCompletedOnboarding` is read ~20 lines before the existing-user flag is set, so updating users with launch-minimized enabled take the wrong branch.

**What the code actually does:** The PR flips the `@AppStorage("hasCompletedOnboarding")` default from `true` to `false` (hyperwhisperApp.swift:75, diff hunk at lines 70-81). In the same sequential launch block, line 572 evaluates `if launchMinimized && hasCompletedOnboarding` (hides the main window), while the existing-user marker `hasCompletedOnboarding = true` is not written until line 605, inside the `else` branch of `if onboardingPending` at :598. For an upgrading user the key was never persisted (the old code defaulted it to `true` and onboarding was disabled, so `completeOnboarding` never ran), so at line 572 it reads `false` and the window is NOT hidden despite `launchMinimized` being on.

**Corrected statement:** Confirmed as written; scope is one launch — the wrong branch is taken only on the first launch after updating, because line 605 persists the flag and subsequent launches minimize correctly.


### `app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift:397` — ✅ **CONFIRMED**

**Original claim:** Three new model-pill localization keys exist only in Base.lproj and are missing from all 40 localized .strings files; String.localized has no value: default, so non-English users see raw keys.

**What the code actually does:** `grep -rl '"onboarding.model.pill.parakeet"'` (and `.whisper`, `.recommended`) over app/macos/hyperwhisper/Localizations/ matches only Base.lproj/Localizable.strings, while sibling new keys such as `onboarding.model.metric.speed` and `onboarding.welcome.title` match 40 files each. A Base-vs-fr key diff lists exactly these three as the only onboarding keys missing. The keys are consumed at OnboardingSourceViews.swift:396-401 via `pill(_:tint:)` (:442-450), whose body is `Text(key.localized.uppercased())`; `String.localized` is `NSLocalizedString(self, comment: "")` (Extensions/LocalizedString.swift:16-18) with no `value:` argument, so a missing key returns the key itself. Non-English users see "ONBOARDING.MODEL.PILL.PARAKEET" etc. in the pill capsules.


### `app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift:622` — ✅ **CONFIRMED**

**Original claim:** testAPIKey() writes the API key to the Keychain BEFORE validating it and never rolls back on .unauthorized, leaving a dead key stored.

**What the code actually does:** OnboardingSourceViews.swift:619-630: `testAPIKey()` calls `settingsManager.apiKeys.setAPIKey(apiKeyInput, for: selectedProvider)` and `cloudHealth.registerAPIKeyChange(...)` synchronously, then starts the `Task` that awaits `cloudHealth.ensureHealthy(selectedProvider)`; the Task only assigns `providerTestHealth`, `keyValidated`, `isTestingKey` — there is no delete/restore on a non-`.healthy` result. `setAPIKey` routes to APIKeySettingsManager.swift:213-231, whose per-provider properties (e.g. `openAIAPIKey`, :47-51) persist via `saveAPIKeyToKeychain(...)` in `didSet`. The in-code comment at :617-618 calls this "harmless", but it is not: a returning user who mistypes over a previously working key has the good key overwritten in the Keychain with no undo, and `.unreachable` (offline) persists a key that was never validated at all.


### `app/macos/hyperwhisper/Views/OnboardingView.swift:381` — ✅ **CONFIRMED**

**Original claim:** appState.lastTranscription also carries raw English error strings (4 call sites), so the onboarding success view can display "Error: ..." as if it were a transcript.

**What the code actually does:** OnboardingView.swift:373 gates the panel on `!appState.lastTranscription.isEmpty` and :381 renders `Text(appState.lastTranscription)` under the "onboarding.try.transcript.heading" label with a green "never pasted" badge — with no `hasPrefix("Error:")` check (contrast RecordingDialog.swift:34-35, which does check that prefix). Error strings are assigned at RecordingTranscriptionFlow+StopRecording.swift:277 (`"Error: Audio file could not be read"`) and :334, and RecordingTranscriptionFlow+ErrorHandling.swift:37 (`"Error: Audio file missing for retry"`), :75 and :355 (`"Error: \(error.localizedDescription)"`). None of these paths consult `TextDeliveryGate.isSuppressed` or the `.onboarding` trigger — the PR's suppression (StopRecording.swift:489-512) only guards the *success* delivery branch.

**Corrected statement:** Substance confirmed, two details off: there are 5 `"Error: ..."` assignment sites, not 4 (StopRecording:277, :334; ErrorHandling:37, :75, :355); and the affected view is the step-6 "give it a try" transcript panel, not the step-7 completion view. Also only the `"Error: "` prefix and the two hardcoded messages are raw English — the interpolated `error.localizedDescription` bodies are localized (TranscriptionError.swift:43+).


---

PR #18, opened 2026-07-02
macOS: fix stop→paste freeze (serial Core Data writer + record-start prewarm + collapsed WAV wait)

ray-amjad:
## Summary

Fixes the stop→paste freeze (up to ~10s observed) in the macOS app. It was **three independent stalls coinciding**; this fixes them without a Core Data model-version bump. Ports the *spirit* of private-repo PR #781 after an independent audit of this repo.

## What changed

**WS1 — Serial background Core Data writer (the confirmed freezer).**
A single long-lived background context (`PersistenceController.performWrite`) now handles every stop-flow write. This removes the ~4.7s of main-thread saves + `processPendingChanges()` + `refreshAllObjects()` that ran right before paste, and serializes writes so rapid stop presses can't create duplicate transcripts. New domain methods: `createProcessingTranscriptInBackground` (folds in the VAD trimmed path so the separate `setTrimmedAudioPath` write is gone), `createFailedTranscriptInBackground`, `updateTranscriptWithTranscriptionInBackground`, `markTranscriptFailedInBackground`, `updateRecordingSessionOnStopInBackground`, `updateTranscriptAudioFilePathInBackground`, `deleteRecordingSessionInBackground`. Only value types + `NSManagedObjectID` cross context boundaries. Paste is now dispatched **before** the completed-status write. View-context re-faulting is debounced off the hot path. Session creation also moved onto the shared writer. The `@MainActor` Core Data APIs are untouched (Streaming / FileTranscription / Retry / TranscriptActionHandler still use them).

**WS2 — Record-start local-model prewarm.**
`handleStartRecording` fires a non-blocking, attempt-id-guarded `prepareModel` + `prepareLocalRuntime` for the exact resolved local mode, closing the residual cold-load window after `ModelResidencyRegistry` eviction or a Quick-Capture pinned mode. Cloud modes skip (they no-op in `prepareModel`).

**WS3 — Collapsed the duplicate WAV wait.**
`stopRecording()` already validates the file exists and is ≥5KB. The downstream `fileWatcher.waitForFirstWrite` + 5×100ms + 20×150ms re-wait is replaced by a single `AVAudioFile`-open readability check (with a 2×50ms retry for the genuinely-async edge). The lifecycle stalled-exit is now a ~400ms wall-time bound instead of the attempt-count arming that yielded ~700–800ms.

**WS4 — Lightweight per-stage timings.**
Per-stage durations (`wav_ready`, `file_check`, `vad_trim`, `create_row`, `transcribe`, `core_data_update`, `paste`) in the completion log line, plus Sentry scope **extras** (which survive `beforeSend`) attached on slow/error paths only — no new always-on Sentry traffic.

## Intentionally not ported from #781
- Full `TimingLog` / `PerfBaseline` observability subsystems (user chose lightweight timings only).
- The `LibWhisper.swift` missing `self.` fix — already correct here.
- The coarse-poll fix — this repo's primary poll is already tiered.

## Verification
- `xcodebuild -project hyperwhisper.xcodeproj -scheme hyperwhisper -configuration Debug build` → **BUILD SUCCEEDED**.
- Grep invariants: no `viewContext.save()` / `performAndWait` / `processPendingChanges` / `refreshAllObjects` reachable from the stop flow; `HyperWhisper.xcdatamodeld` diff empty.
- Runtime pass (record → stop → paste, rapid back-to-back, VAD, M4A flip, forced failure + Retry, too-short discard, streaming unchanged, Quick Capture cold model) to be done via Xcode per repo convention (CLI build verifies compile only).

### DIFF
--- a/app/macos/hyperwhisper/CoreData/PersistenceController.swift
+++ b/app/macos/hyperwhisper/CoreData/PersistenceController.swift
@@ -833,8 +833,328 @@ class PersistenceController: ObservableObject {
         }
     }
     
+    // MARK: - Background Writer (serial)
+
+    /// Long-lived, serial background context that all stop-flow writes funnel
+    /// through. Because every write runs on this single private queue, writes are
+    /// serialized (closing the duplicate-transcript race under rapid stop presses)
+    /// AND kept entirely off the main thread (removing the ~4.7s main-thread Core
+    /// Data stall that froze the stop→paste path).
+    ///
+    /// `automaticallyMergesChangesFromParent = true` means the `viewContext` picks
+    /// up these writes via the standard `NSManagedObjectContextDidSave` merge, so
+    /// HistoryView's existing save-notification observer refreshes with no change.
+    private lazy var writerContext: NSManagedObjectContext = {
+        let context = container.newBackgroundContext()
+        context.name = "HyperWhisper.writer"
+        context.mergePolicy = NSMergeByPropertyObjectTrumpMergePolicy
+        context.automaticallyMergesChangesFromParent = true
+        context.undoManager = nil
+        return context
+    }()
+
+    /// Debounced view-context maintenance task (cancel-previous). Keeps the
+    /// re-faulting benefit off the stop→paste hot path.
+    @MainActor private var viewContextMaintenanceTask: Task<Void, Never>?
+
+    /// Run a write block on the serial background writer context.
+    ///
+    /// The block receives the writer context and performs its mutations; if it
+    /// leaves pending changes they are saved on the writer queue. On save failure
+    /// the context is rolled back (never leaving the long-lived context poisoned)
+    /// and the error is logged + captured. After every write the writer context is
+    /// re-faulted to stay lean, and view-context maintenance is scheduled.
+    ///
+    /// Pass ONLY value types + `NSManagedObjectID` into `block` — never a managed
+    /// object — so nothing crosses a context boundary.
+    @discardableResult
+    func performWrite<T: Sendable>(_ block: @escaping (NSManagedObjectContext) -> T) async -> T {
+        let context = writerContext
+        let result = await context.perform {
+            let value = block(context)
+            if context.hasChanges {
+                do {
+                    try context.save()
+                } catch {
+                    let nsError = error as NSError
+                    AppLogger.logCoreData(.save, error: nsError)
+                    SentryService.capture(
+                        error: nsError,
+                        message: "Core Data background write failed",
+                        tags: ["component": "PersistenceController", "operation": "performWrite"]
+                    )
+                    // Never leave the long-lived writer context poisoned.
+                    context.rollback()
+                }
+                // Re-fault so the long-lived writer doesn't accumulate objects.
+                context.refreshAllObjects()
+            }
+            return value
+        }
+        await MainActor.run { self.scheduleViewContextMaintenance() }
+        return result
+    }
+
+    /// Debounced (cancel-previous) view-context re-faulting. After ~3s of write
+    /// quiescence, if the view context has no pending edits, re-fault it to keep
+    /// it lean without ever touching the stop→paste path.
+    @MainActor
+    private func scheduleViewContextMaintenance() {
+        viewContextMaintenanceTask?.cancel()
+        viewContextMaintenanceTask = Task { @MainActor [weak self] in
+            try? await Task.sleep(nanoseconds: 3_000_000_000)
+            guard !Task.isCancelled, let self else { return }
+            let viewContext = self.container.viewContext
+            guard !viewContext.hasChanges else { return }
+            viewContext.refreshAllObjects()
+        }
+    }
+
+    // MARK: - Background Domain Writes (stop flow)
+
+    /// Create a processing transcript on the serial writer.
+    ///
+    /// Mirrors `createProcessingTranscript`'s idempotency guard, and additionally
+    /// accepts the VAD trimmed path so the separate `setTrimmedAudioPath` write
+    /// collapses into this single create. Returns the permanent object ID (obtained
+    /// before save) so callers can address the row later without crossing contexts.
+    func createProcessingTranscriptInBackground(
+        duration: TimeInterval,
+        mode: String?,
+        audioFilePath: String?,
+        trimmedAudioPath: String?
+    ) async -> NSManagedObjectID? {
+        return await performWrite { context -> NSManagedObjectID? in
+            let normalizedPath = audioFilePath?.trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+
+            var reused: Transcript?
+            if !normalizedPath.isEmpty {
+                let request: NSFetchRequest<Transcript> = Transcript.fetchRequest()
+                request.predicate = NSPredicate(
+                    format: "status == %@ AND audioFilePath == %@",
+                    "processing",
+                    normalizedPath
+                )
+                request.sortDescriptors = [NSSortDescriptor(keyPath: \Transcript.date, ascending: false)]
+
+                if let existing = try? context.fetch(request), let primary = existing.first {
+                    if existing.count > 1 {
+                        for duplicate in existing.dropFirst() {
+                            context.delete(duplicate)
+                        }
+                        AppLogger.coreData.warning(
+                            "Collapsed duplicate processing transcripts for audio path: \(normalizedPath, privacy: .public) (\(existing.count, privacy: .public) -> 1)"
+                        )
+                    }
+
+                    primary.text = "Processing transcription..."
+                    primary.date = Date()
+                    primary.duration = max(primary.duration, duration)
+                    primary.mode = mode
+                    primary.audioFilePath = normalizedPath
+                    primary.setValue("processing", forKey: "status")
+                    reused = primary
+                }
+            }
+
+            let transcript: Transcript
+            if let reused {
+                transcript = reused
+            } else {
+                let created = Transcript(context: context)
+                created.id = UUID()
+                created.text = "Processing transcription..."
+                created.date = Date()
+                created.duration = duration
+                created.mode = mode
+                created.audioFilePath = audioFilePath
+                created.setValue("processing", forKey: "status")
+                transcript = created
+            }
+
+            if let trimmedAudioPath, !trimmedAudioPath.isEmpty {
+                transcript.setValue(trimmedAudioPath, forKey: "trimmedAudioFilePath")
+            }
+
+            try? context.obtainPermanentIDs(for: [transcript])
+            return transcript.objectID
+        }
+    }
+
+    /// Create an already-failed transcript on the serial writer in ONE write
+    /// (replaces the old create + mutate + second save on the error paths).
+    func createFailedTranscriptInBackground(
+        duration: TimeInterval,
+        mode: String?,
+        audioFilePath: String?,
+        failedReason: String,
+        errorText: String
+    ) async -> NSManagedObjectID? {
+        return await performWrite { context -> NSManagedObjectID? in
+            let created = Transcript(context: context)
+            created.id = UUID()
+            created.date = Date()
+            created.duration = duration
+            created.mode = mode
+            created.audioFilePath = audioFilePath
+            created.setValue("failed", forKey: "status")
+            created.setValue(failedReason, forKey: "failedReason")
+            created.text = errorText
+            try? context.obtainPermanentIDs(for: [created])
+            return created.objectID
+        }
+    }
+
+    /// Complete a processing transcript on the serial writer. Same field updates
+    /// and near-now dedup as the `@MainActor` variant, but no `processPendingChanges()`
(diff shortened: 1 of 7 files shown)


app/macos/hyperwhisper/CoreData/PersistenceController.swift
```diff
@@ -833,8 +833,328 @@ class PersistenceController: ObservableObject {
...
+            if let reused {
+                transcript = reused
+            } else {
+                let created = Transcript(context: context)
+                created.id = UUID()
+                created.text = "Processing transcription..."
+                created.date = Date()
+                created.duration = duration
+                created.mode = mode
+                created.audioFilePath = audioFilePath
```
ray-amjad, line 990: **[1/6 · CONFIRMED · data loss] `createProcessingTranscriptInBackground` returns an objectID captured before `performWrite`'s save/rollback.**

The block returns `transcript.objectID` (also `obtainPermanentIDs` swallowed with `try?` at :979), but `performWrite` runs `save()` *after* the block and, on failure, `context.rollback()` — which deletes the just-inserted row while the caller still holds the now-dangling ID.

Failure path: the writer save fails and rolls back, but a non-nil objectID is still returned. The later `updateTranscriptWithTranscriptionInBackground` / `markTranscriptFailedInBackground` `existingObject(with:)` lookups then fail and are silently skipped, so a successfully transcribed-and-pasted result never lands in history — no row at all for that recording.

Fix: return the ID only after a successful save (e.g. have `performWrite`/the domain method surface save success, and return `nil` on rollback), so callers treat a failed create as `nil` and skip downstream updates rather than chasing a dead ID.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+ErrorHandling.swift
```diff
@@ -318,51 +324,61 @@ extension RecordingTranscriptionFlow {
...
+                        appState?.showErrorAlert = false
+                    } else {
+                        appState?.showError(error.localizedDescription)
+
+                        // Store reference to failed transcript for retry button.
+                        // Resolve on the view context now that the writer has persisted
+                        // the failed status (auto-merge has applied it).
+                        if let processingTranscriptID,
+                           let failed = try? PersistenceController.shared.container.viewContext.existingObject(with: processingTranscriptID) as? Transcript {
+                            appState?.lastFailedTranscript = failed
```
ray-amjad: **[2/6 · PLAUSIBLE · retry broken] `lastFailedTranscript` is no longer set unconditionally.**

Old code assigned `appState.lastFailedTranscript = processingTranscript` (a live view-context object) on every non-network failure. New code only sets it when `viewContext.existingObject(with: processingTranscriptID)` succeeds right after the writer save.

If the create-row write rolled back (see finding 1) but a stale objectID was still returned, `existingObject(with:)` returns nil and `lastFailedTranscript` never gets set — the error alert's Retry button then has nothing to act on and the user cannot retry. Tied to finding 1; fixing the objectID contract closes most of this, but consider a fallback so Retry still works when the row is gone.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
```diff
@@ -239,155 +249,93 @@ extension RecordingTranscriptionFlow {
...
+        // FILE READINESS CHECK (WS3):
+        // `recordingLifecycle.stopRecording()` already ran `waitForRawFileReady`,
+        // which guarantees the file exists and is ≥5KB before this flow proceeds.
+        // The old downstream `fileWatcher.waitForFirstWrite` + 5×100ms + 20×150ms
+        // re-waited on that same validated file, adding up to ~1.2s. Collapse it to a
+        // single open/readability test plus one short retry for the genuinely-async
+        // edge (the file was validated moments ago, so this virtually always passes
+        // on the first check).
+        let fileCheckStart = Date()
+        var isReadable = isAudioFileReadable(audioURL)
```
ray-amjad, line 261: **[3/6 · PLAUSIBLE · can discard valid audio] Collapsed readability check is both stricter and much shorter.**

`isAudioFileReadable` now additionally requires `(try? AVAudioFile(forReading: url)) != nil`, and the retry loop is only `for _ in 1...2` with 50ms sleeps (~100ms), replacing ~1.2s of prior readability polling. A freshly finalized file that is momentarily not openable as a decodable audio file (I/O latency, metadata still settling) fails, and the flow marks it `Error: Audio file could not be read`, forcing a manual retry where the old wait would have transcribed it.

Note: a sibling candidate was refuted because `stopRecording()` already gates through `validateRawWAV` (`AVAudioFile(forReading:)`) before the rename — so in the normal path the file is already proven decodable. Worth confirming that gate always precedes this check on every branch (incl. streaming / conversion) before relying on it; if so this is low-risk, otherwise widen the retry window.

app/macos/hyperwhisper/Managers/AudioRecording/Recording/RecordingLifecycle.swift
```diff
@@ -1050,18 +1059,12 @@ class RecordingLifecycle {
...
-                            AppLogger.audio.warning("Raw audio file size stalled at \(fileSize) bytes for \(stalledChecks) checks, giving up after \(attempt) attempts (\(totalWaitMs)ms)")
-                            break
-                        }
-                    } else {
-                        stalledChecks = 0
-                    }
+                // Stalled-size early exit (wall-time bound): if the file has a
+                // nonzero but sub-threshold size that hasn't grown, and we've
+                // already waited past ~400ms, it isn't going to grow — bail now.
+                if fileSize > 0, fileSize == lastSize, totalWaitMs >= stalledExitWallMs {
```
ray-amjad: **[4/6 · PLAUSIBLE · can discard valid audio] Stalled-size early-exit is far more aggressive than before.**

New condition `fileSize > 0, fileSize == lastSize, totalWaitMs >= 400` bails after a *single* unchanged poll past ~400ms (≈attempt 13), vs the old rule of 5 consecutive unchanged checks only after attempt 10 (which tolerated toward ~3.75s). On a slow/contended volume — or the buffer-not-yet-flushed case this function's own comment calls out — the WAV can sit at just its ~4KB header for several hundred ms before data flushes; the new logic exits early, `waitForRawFileReady` returns false, and `RecordingLifecycle` throws `recordingTooShort`, losing a real recording. Consider requiring ≥2 consecutive stable reads, or gating the early-exit on a minimum plausible size.

app/macos/hyperwhisper/CoreData/PersistenceController.swift
```diff
@@ -833,8 +833,328 @@ class PersistenceController: ObservableObject {
...
+    /// Long-lived, serial background context that all stop-flow writes funnel
+    /// through. Because every write runs on this single private queue, writes are
+    /// serialized (closing the duplicate-transcript race under rapid stop presses)
+    /// AND kept entirely off the main thread (removing the ~4.7s main-thread Core
+    /// Data stall that froze the stop→paste path).
+    ///
+    /// `automaticallyMergesChangesFromParent = true` means the `viewContext` picks
+    /// up these writes via the standard `NSManagedObjectContextDidSave` merge, so
+    /// HistoryView's existing save-notification observer refreshes with no change.
+    private lazy var writerContext: NSManagedObjectContext = {
```
ray-amjad: **[5/6 · PLAUSIBLE · concurrency] `writerContext` is a non-atomic `lazy var` reached from concurrent callers.**

`performWrite` is `nonisolated async` and is invoked both from `@MainActor` callers (`RecordingSessionManager`) and from background tasks (`performBackgroundWAVToM4AConversion`). Swift `lazy` stored-property init on a class is not atomic, so two concurrent first-accesses can create two `NSManagedObjectContext`s — defeating the single-serial-queue guarantee and reopening the duplicate-processing-transcript race this change was meant to close. Initialize `writerContext` eagerly at init (or behind a guaranteed-serial accessor) so exactly one writer context exists.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+ErrorHandling.swift
```diff
@@ -318,51 +324,61 @@ extension RecordingTranscriptionFlow {
...
-                    appState?.lastFailedTranscript = processingTranscript
+            Task {
+                if let processingTranscriptID {
+                    await PersistenceController.shared.markTranscriptFailedInBackground(
+                        transcriptID: processingTranscriptID,
+                        failedReason: error.localizedDescription,
+                        errorText: "Transcription failed: \(error.localizedDescription)"
+                    )
+                }
+                await MainActor.run {
```
ray-amjad, line 351: **[6/6 · CONFIRMED · latency asymmetry] Error path awaits the Core Data write *before* showing the error UI.**

`await markTranscriptFailedInBackground(...)` runs before the `MainActor.run` block that shows the error, sets `.idle`, disables the shortcut, and ends power activity — the opposite of the success path (paste/UI first, persist after). When the serial writer queue is backed up (rapid stop presses), the error dialog and idle state are delayed until every queued write drains — re-introducing exactly the write-coupling this PR removes on the success path. Show the error UI first, then persist (fire-and-forget the writer call), mirroring the success path.

app/macos/hyperwhisper/CoreData/PersistenceController.swift
ray-amjad: Fixed in e7b4017. `performWrite` is now a wrapper over a private save-reporting core, and both transcript creates (plus `RecordingSessionManager.createRecordingSession`) go through a new `performWriteRequiringSave` that returns nil unless the save actually persisted. `obtainPermanentIDs` failures now log and return nil instead of silently handing back a temporary ID. Session resolution also switched to `existingObject(with:)`.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+ErrorHandling.swift
ray-amjad: Moot in practice, but hardened anyway (e7b4017). `lastFailedTranscript` has zero readers — the Retry button uses `pendingRetryAudioPath` + `retryTranscriptionFromPendingFile()` (RecordingDialog.swift), so the "retry breaks" scenario can't occur. The assignment is kept for the existing AppState contract and is now honest: finding 1's fix means the ID only exists when the row persisted, and the resolve happens after the writer await (auto-merge applied) via `existingObject(with:)`. Added a comment noting the property is currently vestigial.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
ray-amjad: Partially addressed in e7b4017. The strict `AVAudioFile` check stays — transcription needs decodability anyway, and every batch-flow `audioURL` already passed `validateRawWAV` (a full AVAudioFile decode) upstream in RecordingLifecycle before rename/conversion, so the first check passes at ~0ms cost in practice. What was widened is the failure-path-only retry window: 4×100ms (~400ms worst case) instead of 2×50ms, as insurance for a writer flush landing just after stop.

app/macos/hyperwhisper/Managers/AudioRecording/Recording/RecordingLifecycle.swift
ray-amjad: Fixed in e7b4017. The stalled-size early exit now requires 2 consecutive stable reads (plus the existing 400ms wall-time floor) before bailing, so a single not-yet-flushed poll no longer aborts the wait. Worst-case exit moves from ~450ms to ~650ms — still far under the old ~3.75s full wait.

app/macos/hyperwhisper/CoreData/PersistenceController.swift
ray-amjad: Fixed in e7b4017. `writerContext` is now a `let` assigned eagerly at the end of `init` (after `loadPersistentStores` + viewContext config), so exactly one writer context ever exists and the serial-write guarantee holds regardless of which thread touches it first.

app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+ErrorHandling.swift
ray-amjad: Fixed in e7b4017. Both branches of `handleTranscriptionError` now mirror the success path: the `MainActor.run` UI block (showError / idle state / shortcut disable / power activity) runs first, then the background failed-status write, then — generic non-network branch only — a second small `MainActor.run` resolves `lastFailedTranscript` after the writer await so auto-merge has applied the failed status.

PR #39, opened 2026-07-10
Unify LLM post-processing completion policy in shared-core-rs

ray-amjad:
## What this PR does

Introduces the invariant **"incomplete LLM output must never replace a transcript"** for post-processing — and implements it **once, in the shared Rust core**, instead of three times in three languages. macOS and Windows call the core via uniffi; the cloud backend is chained to the same semantics through a conformance-vector JSON consumed by both the Rust and bun test suites.

### The policy (shared-core-rs `hw_text::completion`)

Two steps, exposed over FFI as `normalizeTermination` / `evaluateCompletion` / `evaluateLlmResponseJson`:

1. **Acceptance gate = termination state, not the wrapper.**
   - OpenAI `finish_reason`: `stop` → Complete, `length` → OutputLimit, other → Incomplete, missing → **Unspecified (proceeds)**
   - Anthropic `stop_reason`: `end_turn`/`stop_sequence` → Complete, `max_tokens` → OutputLimit, other → Incomplete, missing → Unspecified
   - `OutputLimit` / `Incomplete` / `Malformed` → **reject, keep the raw transcript**. Missing metadata proceeds — custom OpenAI-compatible servers routinely omit `finish_reason`, and rejecting them would break every custom-endpoint user.
2. **Text handling = lenient marker removal everywhere.** If a `<<CLEANED>>` variant is present → extract the wrapped content; otherwise strip stray markers, reject leaked prompt scaffolding (`contains_prompt_markers`) and empty results. This is the existing `hw-text::strip_wrapper_markers` behavior, reused (shared marker tables, not duplicated), with a distinguishable failure taxonomy for logs: `None | OutputLimit | IncompleteResponse | MalformedResponse | PromptLeakage | EmptyCleanedText`.

`shared-conformance/completion-vectors.json` encodes every mapping + wrapper edge case once; `hw-text/tests/completion_vectors.rs` (Rust) and `hyperwhisper-cloud/src/lib/llm-completion.test.ts` (bun) both run all cases. A new `shared-core-tests.yml` workflow runs `cargo test` + clippy on PRs touching `shared-core-rs/**` or `shared-conformance/**` (previously honor-system).

### Platform rewires

- **macOS**: `LocalLlmGenerationPolicy` deleted; all four paths (non-streaming BYO, local streaming, custom endpoint, HyperWhisper Cloud) go through the core. This **closes a real gap**: BYO cloud providers had no enforcement at all (the old guard was `provider == .localLLM` and Anthropic's `stop_reason` was never parsed). Malformed responses stay retryable; state rejections are terminal (local → banner, remote → log + raw transcript). Cloud paths (both call sites) now trust the backend's server-side enforcement instead of re-extracting markers.
- **Windows**: the C# policy types/helpers are deleted; transports return raw response JSON and `ProcessAsync` calls the generated `EvaluateLlmResponseJson` (`EvaluateCompletion(…, Unspecified)` for in-process LLamaSharp). The orphaned `ExtractCleanedText*`/marker-array cluster in `PromptBuilder` is removed. Smoke tests drive the FFI functions directly on the real DLL — doubling as the uniffi checksum-drift gate.
- **Cloud**: `getLLMCompletionStatus` gains an `unspecified` state (missing metadata proceeds); the strict `extractCompleteCleanedText` is dropped and text handling reverts to main's lenient `stripCleanMarkers` + `containsPromptLeakage`, with an explicit empty-after-strip fallback. **No runtime wire-contract change.**

### Semantics changes vs the first revision of this PR

- **Lenient, not strict, wrapper handling**: the earlier revision required a byte-exact `<<CLEANED>>…<<END>>` wrapper (Windows/cloud). Under load, models that produce a complete cleaned transcript but skip or typo the wrapper would have silently lost their output; lenient matching (marker variants, markerless passthrough with a prompt-leakage guard) restores main's behavior while keeping the new termination gate.
- **Missing `finish_reason` proceeds** instead of failing closed — fixes the custom-endpoint regression.

### Kept from the first revision as-is

Windows transport dedup (one OpenAI-compatible transport + endpoint enum), token/context raises (8,192 output ceiling, bigger llama-server context), `--no-context-shift`, tokenizer-based prompt-budget rejection, and the Windows cloud-path trust-normalized-response change.

### Known residual risk

Windows' in-process LLamaSharp surfaces no finish reason (always `Unspecified`), so a truncated, markerless local response could pass the lenient path. Mitigated by the 8,192-token output ceiling; macOS local (llama-server) does report `finish_reason` and is fully covered.

### Known drift, explicitly out of scope (phase 2)

- TS `containsPromptLeakage` and Rust `contains_prompt_markers` use different marker sets (TS has `--TRANSCRIPT--`/`<INSTRUCTIONS>`; Rust has more section tags). Vectors only use markers present in both.
- ~~Rust checks leakage only on the unwrapped path; TS checks post-strip regardless of wrapper presence.~~ **Fixed in the post-review commits** — Rust now checks leakage on both paths (see below).
- Provider endpoint/quirks catalog as shared data; token-limit constants exported from the core; moving request building / HTTP transports into Rust; non-post-process `stripCleanMarkers` call sites.

### Verification

- `shared-core-rs`: `cargo test` (all crates incl. 23 conformance vectors + FFI JSON tests) + `cargo clippy --all-targets -D warnings` — clean.
- `hyperwhisper-cloud`: `bun run typecheck` + `bun test src` — 129 pass, 0 fail (vector suite asserts accepted/text/failure per case).
- macOS: `xcodebuild build` + `build-for-testing` succeed (CLI verifies compile per repo convention; the Swift test target and a manual end-to-end pass still need an Xcode-signed run).
- Windows: managed build compiles cleanly on macOS; the full smoke run (incl. the FFI checksum handshake) happens in `windows-release.yml` CI.
- Bindings freshness: regenerated via `build-bindings.sh`, copies diffed against `shared-core-rs/bindings/` (in sync); `libhyperwhisper_core.a` rebuilt and committed once.

### Post-review fixes

A code review of this PR produced 8 verified findings (3 correctness, 5 cleanups), fixed in four follow-up commits:

- `fix(core)`: reject prompt leakage inside `<<CLEANED>>…<<END>>` wrappers on the Rust path too (was unwrapped-path-only) + new `leakage-inside-wrapper-rejected` conformance vector.
- `fix(cloud)`: bill a successful alternate-provider retry immediately (cost was dropped if a later step threw); present-but-empty content (`message.content: ""`) is now a graceful raw-transcript fallback instead of a billed 500; both route paths consume a new shared `evaluateCompletionResponse` helper mirroring the core's `evaluate_completion`.
- `refactor(windows)`: collapse six identical OpenAI-compatible switch arms; `TestEndpointAsync` reuses `BuildOpenAIRequestJson`; prompt-budget guard runs before KV-cache allocation; smoke tests dedupe + a wrapped-leakage FFI check on the real DLL.
- `chore(macos)`: rebuilt `libhyperwhisper_core.a` (bindings diffed clean — no API change) + a Swift spot check for the new vector.

https://claude.ai/code/session_019mPeULZ7CP593WF5S83Hm5


ray-amjad:
## Code review summary + fixes

A full review of this PR produced **8 verified findings — 3 correctness bugs and 5 cleanups** — all fixed in the four follow-up commits now on this branch.

### Correctness

| # | Finding | Fixed in |
|---|---------|----------|
| 1 | **Retry billing gap (cloud):** after a prompt-leakage retry, `costUsd += retryResponse.costUsd` ran only at the end of the retry block — if evaluating the retry response threw, a provider call we paid for was never billed to the user. Provider/model/cost now update immediately after `callWithRetry` returns. | `fix(cloud)` |
| 2 | **Wrapped-leakage skip (Rust core):** `evaluate_completion` checked `contains_prompt_markers` only on the markerless path, so leaked scaffolding wrapped in `<<CLEANED>>…<<END>>` (e.g. an echoed `<APPLICATION_CONTEXT>` block) was accepted verbatim on macOS/Windows. The leakage check now runs on the cleaned text from both paths, matching the cloud route; covered by the new `leakage-inside-wrapper-rejected` conformance vector, a Windows FFI smoke check, and a Swift spot check. | `fix(core)` |
| 3 | **Empty content billed as a 500 (cloud):** `message.content: ""` with `finish_reason: "stop"` threw in `extractCorrectedText` and hit the `extract_failed` bill+500 path, while the conformance vector expects a graceful `empty_cleaned_text` rejection. Extraction now distinguishes present-but-empty text (returns `''` → raw-transcript fallback, billed once) from a response with no text field anywhere (still a 500). | `fix(cloud)` |

### Cleanups

- **Shared policy helper (cloud):** the gate → extract → leakage → empty cascade was inlined twice in the route and mirrored a third time in the vector test. It now lives once in `evaluateCompletionResponse` (the TS analog of the core's `evaluate_completion`), and the test's `SKIP_FAILURE_ASSERTION` escape hatch is gone — `failure` is asserted for every vector. — `fix(cloud)`
- **Six identical switch arms (Windows):** the OpenAI-compatible providers in `PostProcessingService` collapse into one arm + an explicit `MapToOpenAICompatibleProvider` that throws on non-compat providers. — `refactor(windows)`
- **Endpoint test drift (Windows):** `TestEndpointAsync` hand-built a request (`max_tokens: 10`, `temperature: 0.0`, user-only) that diverged from real traffic; it now reuses `BuildOpenAIRequestJson` for exact structural parity. — `refactor(windows)`
- **Guard after allocation (Windows):** `LocalLlmService` allocated the 16k-token KV cache before checking the prompt token budget; the guard only needs the tokenizer, so it now runs first. — `refactor(windows)`
- **Duplicate smoke test (Windows):** the second "uncapped request JSON" check was byte-for-byte redundant and is removed (replaced by the new wrapped-leakage FFI check). — `refactor(windows)`

### Refuted during review

- *macOS oversized-prompt "data loss"* — a candidate finding claimed an over-budget local prompt could drop the transcript. Verification showed it falls back to the raw transcript safely; not a bug.

### Deliberately deferred

- Swift failure→reason log-string duplication — log-only strings, not user-facing; not worth a helper yet.
- Unifying the 8,192 output-token ceiling as a cross-language constant exported from the core — explicitly phase-2 scope, along with the TS-vs-Rust leakage marker-set reconciliation (still listed under "Known drift").

Verification: `cargo test` + clippy clean (23 vectors), cloud typecheck + 129 tests pass, `xcodebuild build`/`build-for-testing` succeed, Windows managed build compiles (full smoke run happens in Windows CI with a fresh DLL). Bindings regenerated and diffed clean against the committed copies — no API change.

https://claude.ai/code/session_019mPeULZ7CP593WF5S83Hm5

PR #43, opened 2026-07-17
fix(windows): resolve capture device by ordinal position, not fragile name matching

ray-amjad:
Fixes a confirmed subclass of **HYPERWHISPER-PA** (Windows transcription no-speech diagnostic — 60 users, 368 events), restores dropped diagnostic telemetry for the rest, and promotes key fields to Sentry tags for future segmentation.

## Verified against live Sentry data (by Ray — 90d query, all 368 events / 60 users)

I don't have Sentry API/login access in this sandbox, so all code-level investigation below was verification-by-tracing until Ray ran the aggregate query this needed. Two clean populations:

- **`backend_no_speech_detected=true`: 230 events, 48 users** — backend and provider agree there's no speech. The audio itself is the problem (genuinely silent/too quiet). This population is **not** addressed by this PR — it needs a gain/mic-level UX fix (e.g. a chronic-low-RMS warning, input-gain normalization before upload), tracked as follow-up.
- **`backend_no_speech_detected=false`: 138 events, 25 users (~38% of events)** — backend saw speech-ish audio but the provider still returned no-speech. This is the mismatch class this PR targets, and matches the sample event Ray pulled: 1.7s clip, non-silent ratio 0.226, peak -25.5dBFS, RMS -40.6dBFS, 16kHz mono WAV, device `"Microphone (High Definition Aud…"`.

**Verdict (Ray):** scope this PR as the fix for the ~38%/25-user mismatch population; the majority "agreeing no-speech" population needs a separate gain/UX pass and is out of scope here.

## What's in this PR (3 commits)

**1. Ordinal-position device resolution** (`AudioRecorderService.ResolveCaptureDevice`) — the plausible root cause of the mismatch population. Mic-volume boost is supposed to raise a low mic volume to 90% on "the device actually being recorded," resolved by matching NAudio's truncated (31-char) WaveIn `ProductName` as a string *prefix* of the Core Audio `FriendlyName`. Driver-reported names aren't guaranteed to share a prefix at all (e.g. `"Microphone (Realtek(R) Audio)"` vs `"Realtek(R) Audio")`, so on any mismatch/ambiguity it silently fell back to the **default communications endpoint** — a different physical device than the one actually selected and recording. Net effect: boost lands on the wrong device, the real (low-volume) capture device stays untouched, producing recordings with genuine but too-quiet speech — exactly the "provider disagrees" shape.

Fix: switched the primary correlation strategy to **ordinal position** — legacy WinMM (which NAudio's `WaveIn` wraps) enumerates active capture devices in the same relative order as the modern Core Audio active-endpoint list, so the Nth WaveIn device is normally the Nth entry of `EnumerateAudioEndPoints(Capture, Active)`. Doesn't depend on name formatting. Old name-matching kept as a secondary fallback if the two lists ever diverge in size.

**2. Provider diagnostics wiring** (`RustCoreMapping.MapTranscriptionError`, `HyperWhisperCloudService`) — Ray's sample event showed `backend_http_status: 0` and `backend_response_latency_ms: 0`, initially read as a possible transport failure being misreported as no-speech. Traced it: it's dropped telemetry, not a transport failure. `HyperWhisperCloudService` already built a `TranscriptionProviderDiagnostics` with the real HTTP status + latency right before throwing on the "200-but-no-speech" path, but only assigned it to a `LastDiagnostics` property, never to the exception itself — and the shared error mapper every provider's error path goes through never had a parameter to accept one. Fixed by threading `providerDiagnostics` through the full mapper and wiring the throw site to pass it. Going forward, these fields carry real data instead of struct defaults.

**3. Tag promotion** — Ray's query hit a wall: `audio_rms_dbfs` and `selected_input_device_name` are event *extra data*, not tags, so Sentry couldn't facet/aggregate on them for further segmentation. Promoted `audio_rms_dbfs` (bucketed to 5dBFS steps as `audio_rms_dbfs_bucket` — a raw float as a tag is ~100% cardinality and defeats faceting; exact value stays in extras) and `selected_input_device_name` to tags, plus a new `capture_device_count` tag (total audio devices enumerated at capture time) to help spot machines with many/ambiguous devices.

## Residual (explicitly out of scope here)

The 230-event/48-user "backend agrees, audio is genuinely bad" population is real and this PR does not reduce it. Candidate fixes discussed: input-gain normalization before upload, a low-RMS warning in the UI, or a local-Whisper retry/fallback when the provider says no-speech but local analysis disagrees (that last one only helps the mismatch population, not this one). Needs its own scoping pass.

## Other gaps found during investigation (not fixed here, flagging for follow-up)

- Streaming-mode no-speech failures (`MainViewModel.cs:1524-1528`) never call `CaptureNoSpeechDiagnostic` at all — that path's failures are currently invisible in Sentry.
- `OpenAIStreamingStrategy.AudioSampleRate` requests 24kHz capture while every other streaming provider uses 16kHz — a device/driver that can't cleanly deliver 24kHz via the legacy MME path has no app-level verification the rate was honored.
- The file-import flow to the local Parakeet/Qwen3 daemon has no visible resample-to-16kHz/mono step (unlike the whisper.net path's `PrepareAudioStream`), so an imported file at the wrong sample rate could plausibly produce empty/garbage results.

## Testing

No unit test project covers these services with live Core Audio/WaveIn/HTTP behavior (only `HyperWhisper.SmokeTests` exists, and this logic needs live hardware/network). I don't have a Windows environment with audio hardware in this sandbox to exercise it live, so this is verified by code review/tracing plus Ray's live Sentry query — not a runtime repro on my end. Flagging that explicitly rather than claiming more than I've checked.

### DIFF
--- a/app/windows/HyperWhisper/Services/AudioRecorderService.cs
+++ b/app/windows/HyperWhisper/Services/AudioRecorderService.cs
@@ -241,15 +241,34 @@ public class AudioRecorderService : IDisposable
 
     /// <summary>
     /// Resolves the Core Audio (MMDevice) capture endpoint that corresponds to a
-    /// NAudio WaveIn index, so volume boost/restore acts on the device actually
-    /// being recorded rather than the system default comms mic.
+    /// NAudio WaveIn index, so volume boost/restore (and mic volume reads) act on
+    /// the device actually being recorded rather than a different mic entirely.
     ///
-    /// NAudio's <c>WaveIn.GetCapabilities(n).ProductName</c> is truncated to 31
-    /// characters (the MME WAVEINCAPS limit), so we match it as a prefix of the
-    /// full Core Audio <c>FriendlyName</c> instead of by equality. A
-    /// <paramref name="deviceNumber"/> of -1 is NAudio's "default device"
-    /// sentinel; when out of range, unmatched, or ambiguous we fall back to the
-    /// default comms endpoint so the feature degrades gracefully instead of throwing.
+    /// PRIMARY STRATEGY - ordinal position:
+    /// Legacy WinMM (<c>waveInGetNumDevs</c>/<c>waveInGetDevCaps</c>, which NAudio's
+    /// <c>WaveIn</c> wraps) enumerates active capture devices in the same relative
+    /// order as the modern Core Audio active-endpoint list on Windows Vista+, since
+    /// both ultimately come from the same device topology. So the Nth WaveIn device
+    /// is normally the Nth entry of <c>EnumerateAudioEndPoints(Capture, Active)</c>.
+    /// This avoids the previous name-matching bug: NAudio's
+    /// <c>WaveIn.GetCapabilities(n).ProductName</c> is truncated to 31 characters
+    /// (the MME WAVEINCAPS limit) and isn't guaranteed to share a common prefix
+    /// with the Core Audio <c>FriendlyName</c> at all (driver-dependent formatting,
+    /// e.g. "Microphone (Realtek(R) Audio)" vs "Realtek(R) Audio") - when it didn't
+    /// match, this method silently fell back to the *default communications
+    /// device*, which may be a completely different physical mic than the one the
+    /// user selected and is actually recording. That meant volume boost could be
+    /// applied to the wrong device while the real (quiet/muted) capture device was
+    /// left untouched, producing near-silent recordings that surface as no-speech.
+    ///
+    /// FALLBACK - truncated-name prefix match:
+    /// Used only if the ordinal position is out of range for the Core Audio list
+    /// (e.g. list sizes momentarily diverge across the two APIs). Kept as a
+    /// best-effort secondary signal rather than removed outright.
+    ///
+    /// A <paramref name="deviceNumber"/> of -1 is NAudio's "default device"
+    /// sentinel; when both strategies fail we fall back to the default comms
+    /// endpoint so the feature degrades gracefully instead of throwing.
     /// </summary>
     private static MMDevice ResolveCaptureDevice(MMDeviceEnumerator enumerator, int deviceNumber)
     {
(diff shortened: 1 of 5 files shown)


ray-amjad:
Update based on the Sentry sample data Ray shared.

## The `backend_http_status: 0` / `latency: 0ms` anomaly — explained, not a transport failure

Traced it to a real (separate) bug, now fixed in the latest commit: `HyperWhisperCloudService.TranscribeAsync` builds a `TranscriptionProviderDiagnostics` with the actual HTTP status + latency right before throwing on the "200-but-no-speech" path — but only assigned it to `LastDiagnostics`, never to the exception itself. `RustCoreMapping.MapTranscriptionError` (the shared error mapper *every* provider's error path goes through) never had a `providerDiagnostics` parameter to accept it, even though `TranscriptionException` already supports carrying one. `TranscriptionOrchestrator`'s own empty-result check *does* thread diagnostics through when it's the one building the exception — but when the provider itself throws (this path), that code never runs; the exception unwinds straight past it.

Net effect: every HyperWhisper-Cloud-routed no-speech event has been reporting `backend_http_status=0`/`backend_response_latency_ms=0`/`backend_no_speech_detected=false` regardless of what actually happened on the wire — those fields were never real telemetry for this path, just the struct defaults. So the sample event's oddity isn't evidence of a transport failure; it's evidence of this wiring gap. Fixed by threading `providerDiagnostics` through `MapTranscriptionError`'s full switch and passing the already-built diagnostics (plus the response status, which also wasn't being passed) into the throw.

**Caveat:** I still don't have Sentry query access, so I can't personally re-verify the "does short+quiet+non-silent hold broadly" question across the 1,271 events — that part of Ray's ask is still open. What I can say: `audio_duration_seconds` / `audio_rms_dbfs` / `audio_non_silent_ratio` come from local WAV analysis (`TranscriptionDiagnosticsService.AnalyzeAudioFile`) and were never affected by this bug, so a historical query on those three is trustworthy as-is. A query on `backend_http_status` against *past* events won't be useful (they're all 0 due to the bug above) — that field only becomes meaningful for events captured after this fix ships.

## Relationship to the device-resolution fix (commit 1)

Both fixes target HYPERWHISPER-PA from different angles and are complementary, not competing theories:
- Commit 1 (ordinal device resolution) addresses a plausible *cause* of quiet-but-real audio: mic-volume boost silently landing on the wrong physical device.
- Commit 2 (this one) restores the *telemetry* needed to actually confirm/refute that theory (and any other) going forward, since the existing data for this specific path was structurally incapable of showing a real HTTP status or latency.

Given the "1.7s clip, RMS -40.6dBFS" sample audio characteristics, I'd guess the low-gain/quiet-mic theory is a reasonable working hypothesis — a genuinely silent/wrong-device recording should trip a different, harder suppression bucket in `TranscriptionDiagnosticsService.ShouldCaptureNoSpeechDiagnostic` — but that's still an inference, not something I've confirmed against the full event set.

PR #44, opened 2026-07-17
fix(macos): populate deviceId/deviceName in crash-recovery stub sessions

ray-amjad:
## Bug
Fixes HYPERWHISPER-V3 and HYPERWHISPER-V4 — same underlying failure, reported twice under different Sentry tags (`PersistenceController` component vs. `coredata` category) because `PersistenceController.save()`'s catch block logs the same error to Sentry via two separate call sites.

## Root cause
`CrashRecoveryManager.synthesizeStubSessionsForUnclaimedWAVs()` (added in `a638801`, shipped in 2.41.0 on July 2) synthesizes a stub `RecordingSession` for any unclaimed `.incomplete_*.wav` file found at launch — recovering recordings left behind by a crashed/force-quit process. The stub sets `id`, `startTime`, `audioFilePath`, `sampleRate`, `channelCount`, and `audioFormat`, but never `deviceId`/`deviceName`.

Those two attributes are non-optional on `RecordingSession` (they always have been — this is not a recent schema change; checked every `.xcdatamodeld` version from introduction through current). A crashed process never reaches the normal record-start path (`RecordingSessionManager.createRecordingSession`) that populates them from the selected input device, so the synthesized stub saves with `deviceId = nil`.

The batch save in `recoverOrphanedRecordings()` is atomic, so `context.save()` throws `NSCocoaErrorDomain` code 1570 ("deviceId is a required value") and the **entire recovery batch silently drops** — on every launch that finds an unclaimed WAV. Matches the Sentry evidence exactly: `deviceId = nil`, `deviceName = nil`, `durationInSeconds = 0`, `status = "recording"`, failure ~2s after launch.

## Fix
Default `deviceId` to `"unknown"` and `deviceName` to `"Unknown Device"` when synthesizing the stub — the real device is unrecoverable at this point, so fall back to a sentinel, consistent with the existing fallback pattern in `RecordingLifecycle.swift` for unresolvable device info.

## Out of scope
An F4 Sentry event shows a separate vocabulary-store Core Data version-hash mismatch (code 134100). Investigated briefly: different persistent store (`Cloud`/CloudKit-backed `Vocabulary`, vs. `Local` for `RecordingSession`), different failure mode (store-open version-hash mismatch vs. save-time validation), no code path shared with this fix. Left untouched — flagging for separate follow-up.

Sentry: HYPERWHISPER-V3, HYPERWHISPER-V4

---

### DIFF
--- a/app/macos/hyperwhisper/Managers/AudioRecording/Recording/CrashRecoveryManager.swift
+++ b/app/macos/hyperwhisper/Managers/AudioRecording/Recording/CrashRecoveryManager.swift
@@ -512,10 +512,10 @@ class CrashRecoveryManager {
             // NSCocoaErrorDomain code 1570 "deviceId is a required value", silently
             // dropping the whole recovery pass (see HYPERWHISPER-V3/V4). The real
             // device is unrecoverable at this point, so fall back to a sentinel —
-            // consistent with the "unknown" fallback already used elsewhere for
-            // unresolvable device info (see RecordingLifecycle.swift).
-            stub.deviceId = "unknown"
-            stub.deviceName = "Unknown Device"
+            // consistent with the "default" fallback already used elsewhere when no
+            // input device resolves (see RecordingLifecycle.swift:459-460).
+            stub.deviceId = "default"
+            stub.deviceName = "audio.device.default".localized
             stubs.append(stub)
 
             AppLogger.audio.info("🩹 Synthesized stub session \(sessionId.uuidString, privacy: .public) for unclaimed incomplete WAV: \(url.lastPathComponent, privacy: .public)")
--- a/app/macos/hyperwhisper/Managers/AudioRecording/Recording/CrashRecoveryManager.swift
+++ b/app/macos/hyperwhisper/Managers/AudioRecording/Recording/CrashRecoveryManager.swift
@@ -503,6 +503,19 @@ class CrashRecoveryManager {
             stub.channelCount = 1
             stub.audioFormat = "WAV PCM 16000Hz 1ch"
             stub.endTime = nil
+            // `deviceId`/`deviceName` are non-optional on RecordingSession (no default
+            // value), but the crashed process that recorded this file never got a
+            // chance to persist them — the row we're synthesizing here never went
+            // through the normal record-start path that populates them from the
+            // selected input device. Leaving them nil causes the batch save in STEP 3
+            // to fail validation for every session in the batch with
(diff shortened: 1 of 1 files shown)


ray-amjad:

**Codex:** clean — "The added sentinel values satisfy the required Core Data attributes for synthesized recovery sessions and prevent the batch save validation failure. No regressions or actionable defects were found."

**Claude (medium tier):** no correctness bugs. 2 non-blocking cleanup findings:

1. `CrashRecoveryManager.swift:514` — The new sentinel pair (`"unknown"` / `"Unknown Device"`) is **inconsistent** with the codebase's actual existing fallback for this field. `RecordingLifecycle.swift:459-460` falls back to `deviceId = "default"` and `deviceName = "audio.device.default".localized` when no device resolves — not `"unknown"`. The added comment's claim of consistency misattributes this precedent (the `"unknown"` string that *does* exist in that file, line 357, is an unrelated local log variable that's never persisted). As written, this introduces a second, differently-named, non-localized sentinel — anything filtering on `deviceId == "default"` or displaying `deviceName` in a non-English locale will treat crash-recovered stub sessions inconsistently.
2. `CrashRecoveryManager.swift:517` — The fix is a narrow, single-site literal fix rather than a Core Data model default or shared construction helper. `deviceId`/`deviceName` are the only two `RecordingSession` attributes without a `defaultValueString` in the model, which is what let this class of bug happen in the first place. Non-blocking, but worth a follow-up: either add a model default, or route all `RecordingSession` construction through one shared factory.

Fixing (1) now — using the existing `"default"` / localized fallback instead of the new literals — since it's a real inconsistency, not just style. (2) noted for a possible follow-up but out of scope for this PR (schema/model changes deserve their own review).

ray-amjad:

**Codex:** clean — "No regressions or actionable defects are evident."

**Claude (medium tier):** no correctness bugs found by any angle, including conventions. Two minor PLAUSIBLE cleanup/altitude notes survive, both explicitly tempered as non-blocking and the review states the diff is safe to merge as-is:
1. The fix is scoped to this one crash-recovery call site rather than a schema-level default — defensible per this repo's CLAUDE.md (schema changes need a full new `.xcdatamodel` version), so not treated as an oversight.
2. The `"default"` / `"audio.device.default".localized` sentinel pair is duplicated (already ~8 places codebase-wide) rather than centralized into a shared constant — pre-existing pattern, no history of drift.

Stopping the review loop here (2 rounds, both reviewers clean of correctness issues). Moving to verification.

PR #2, opened 2026-06-30
macOS: combine license + Cloud credits into one account panel

ray-amjad:
## Summary
Merges the separate **License** and **HyperWhisper Cloud Credits** settings sections into a single, always-visible **"HyperWhisper Cloud"** panel — reflecting the new model where the license key *is* the wallet that holds Cloud credits.

## Changes
- **New `CloudAccountSettingsSection.swift`** replaces `CreditsSettingsSection` + `LicenseSettingsSection`.
- **Licensed view** keeps the existing credits-card style (balance + Account Information) and adds the **license key as a masked, password-style row** with eye-reveal + copy.
- **Single "Purchase More" button** opens the web `…/credits?license_key=…` page — no in-app tier picker; the web page handles amounts and minting/emailing the key.
- **Unlicensed view**: license key-entry block + a "Get Credits" button (also → `/credits`) + restore link.
- **`SettingsView`** collapses the two sidebar entries into one (`license` tag preserved so existing deep links still resolve); removes the now-unneeded deactivate→General redirect.
- Adds `settings.cloud.*` strings to `Base.lproj`.

## Notes
- Dropped the old **trial usage cards** (daily-seconds / model-download counters) from this panel, since the open-source transition removes those local limits anyway.
- New `Base.lproj` keys still need syncing to the 39 sibling locales (`localisation-syncer`) before release.
- Windows app to follow with the same design.

## Test plan
- `xcodebuild -scheme hyperwhisper -configuration Debug build` → **BUILD SUCCEEDED**, zero Swift errors.
- Run via Xcode (Cmd+R) and verify the combined panel: licensed (balance + masked key + Purchase More → /credits) and unlicensed (key entry + Get Credits) states.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1, opened 2026-06-30
Better control over punctuation & automatic formatting in dictation


ray-amjad:
Two related pieces of feedback around how dictated text gets punctuated and formatted:

## 1. Comma placement around direct address
Greetings like "Hi Ray," are sometimes rendered as "Hi, Ray." — the comma lands in the wrong spot (and a period gets used instead of a comma). It would be great to either improve this heuristic or give users a way to specify their preferred convention.

## 2. Automatic formatting (newlines / paragraphs)
Especially for emails, output would benefit from automatic structure — paragraph breaks, new lines, etc. — similar to how some competing tools format dictation. Users should also be able to dictate this structure explicitly (e.g. saying "new line") and have it honored.

Both could plausibly be addressed via post-processing / mode prompts.

PR #45, opened 2026-07-17
Fix macOS main-thread hang during streaming transcription (HYPERWHISPER-F7)

ray-amjad:
## Problem

Sentry HYPERWHISPER-F7 — macOS app hang ≥10s on the main thread, top issue by impact (2,208 occurrences, 67 users, long-standing since 2026-01-17, still firing on hyperwhisper@2.41.0 build 110).

Hang stack points at a SwiftUI observation-driven graph re-evaluation of the window's root view during a scroll-view-heavy update:

```
NSRunLoop.flushObservers → GraphDelegate.beginTransaction → AppGraph.updateGraph →
AppKitWindowController.updateRootHost/updateRootView → GraphHost.preferenceValues →
AG::Graph::update_attribute → TransformScrollStorageEnvironment.updateValue →
ObservationCenter.invalidate
```

## Root cause

`AppState.streamingText` was `@Published` directly on the shared `AppState` object, which is injected via `@EnvironmentObject` into `MainAppView` (the window's actual root view), the menu bar content, `RecordingDialog`, and `StreamingPreviewBubble`.

During a streaming transcription session, interim/final deltas arrive multiple times per second from the active provider (Deepgram / ElevenLabs / xAI / HyperWhisper Cloud / on-device Parakeet / Nemotron), each replacing the string wholesale. Because SwiftUI's `@EnvironmentObject` change tracking is coarse-grained — any `@Published` mutation invalidates *every* view holding the object, not just views reading the changed property — each delta forced a synchronous re-evaluation of the whole main window's view graph (NavigationSplitView, sidebar ScrollView, active tab like History's List), matching the hang stack's graph-update/scroll-storage/observation-invalidate frames. Under memory pressure (~120MB free on the sampled device), a burst of these forced full-window updates plausibly compounds into a ≥10s stall.

The codebase already diagnosed and fixed this exact bug class once before, for a different high-frequency property: `RecordingLiveMetrics` isolates `AudioRecordingManager`'s 30 FPS `audioLevel`/`recordingDuration` off `AppState` for the same reason. `streamingText` was the remaining high-frequency `@Published` value left on the shared object.

## Fix

- New `StreamingLiveText`: a small `@MainActor` `ObservableObject` (mirrors `RecordingLiveMetrics`) holding just the streaming preview text.
- `AppState.streamingText` becomes a computed passthrough (`get`/`set`) to `liveStreamingText.text`, so every existing read/write call site across the streaming pipeline keeps working unchanged — but mutations no longer fire `AppState.objectWillChange`.
- `RecordingDialog` and `StreamingPreviewBubble` (the only two views that actually render the streaming preview) inject `StreamingLiveText` via `@EnvironmentObject` and read from it directly, so they stay fully reactive.
- `RecordingWindowManager` / `StreamingPreviewWindowManager` inject `.environmentObject(appState.liveStreamingText)` at the two window-construction sites, alongside the existing `liveMetrics` injection.

## Caveats

- Sentry app-hang stacks group broadly and this stack is generic SwiftUI/AttributeGraph machinery rather than a literal 1:1 code match. This is the most plausible root cause given the evidence — it matches the exact bug class already fixed once for `audioLevel`, and matches "large payload replacing a `@Published` property forcing full root-level invalidation" — but it's a best-judgment call, not a certainty. Worth sampling a few more Sentry events to confirm before merge if there's doubt.
- No Xcode/Swift toolchain available in this sandbox, so the change could not be compiled here — reviewed manually against all call sites instead. Recommend an `xcodebuild` pass before merge.

Fixes HYPERWHISPER-F7.

---


ray-amjad:

**Codex (`codex review`):** clean. No P1/P2 findings — "The new observable object is consistently injected into the active recording and preview window paths, while existing streaming pipeline call sites remain compatible through the AppState passthrough. No actionable regressions were identified."

**Claude (medium tier, 8 finder angles + verify pass):** 6 findings, none are regressions this PR introduces into a live code path. Two are worth fixing now, the rest are pre-existing/out-of-scope follow-ups:

1. **`MainAppView.swift:274` — `openRecordingDialogWindow()` doesn't inject `liveStreamingText`.** `RecordingDialog` now hard-requires `StreamingLiveText` via `@EnvironmentObject`. This method is currently dead code (zero callers, confirmed via grep) so it's not a live bug today, but if it's ever wired up it fatal-crashes on missing environment object. **Fixing** — cheap defensive injection.
2. **`RecordingDialog.swift:324,353,435` — mixes `appState.streamingText` (non-reactive passthrough) and `liveStreamingText.text` (reactive) for the same value in one file.** No bug today (these are one-shot imperative reads/writes in event handlers, not body/`.onChange` reads), but a future contributor could add a reactive read on `appState.streamingText` believing it's still live and get silently stale UI. **Fixing** — switch to `liveStreamingText.text` for consistency.
3. `MainAppView.swift:27` — `isStreaming`/`streamingConnectionState`/`showStreamingPreview` remain plain `@Published` on `AppState` and still invalidate the main window on session-lifecycle transitions (not per-delta, so unlikely to reproduce the full 10s hang) — mitigates rather than eliminates the invalidation-fan-out mechanism. Noted as a follow-up, not fixing in this PR.
4. `StreamingLiveText.swift` duplicates `RecordingLiveMetrics`'s structure — real but non-trivial cleanup opportunity (no generic wrapper is a clean win given `@EnvironmentObject` dispatches by concrete type). Follow-up.
5. `AppState` still has 34 `@Published` fields; migrating to `@Observable` (macOS 14+, already the deployment target) would fix this class of bug generically instead of one-field-at-a-time carve-outs. Out of scope for this hotfix — flagging as a larger follow-up.
6. `RecordingDialog.swift:1338` — `#Preview` block is missing `.environmentObject(TranscriptionPipeline())`, pre-existing gap unrelated to this diff (dev-tooling only, no runtime impact). Not fixing.

Applying fixes for #1 and #2 now.

ray-amjad:

Both clean.

- **Codex:** "The added environment-object injections match RecordingDialog's dependencies, and the direct liveStreamingText accesses preserve existing behavior while using the reactive source of truth. No regressions were identified."
- **Claude (medium, 8 finder angles + verify):** 0 findings survived verification. Both round-1 fixes (env-object injection in `openRecordingDialogWindow`, and `RecordingDialog.swift`'s switch to `liveStreamingText.text`) independently confirmed correct — no missed injection sites, no orphaned reactivity, no threading/actor-isolation issues.

Stopping the review loop here (clean on both models). Remaining round-1 items (architectural `@Observable` migration, `StreamingLiveText`/`RecordingLiveMetrics` duplication, the residual lower-frequency `AppState` invalidation, the pre-existing `#Preview` gap) are still noted above as deliberate out-of-scope follow-ups, not fixed in this PR.

app/macos/hyperwhisper/Views/RecordingDialog.swift
```diff
@@ -1336,8 +1338,11 @@ struct VisualEffectBackground: NSViewRepresentable {
...
 
 #Preview {
+    let previewAppState = AppState()
     RecordingDialog(isPresented: .constant(true))
         .environmentObject(AudioRecordingManager())
-        .environmentObject(AppState())
+        .environmentObject(previewAppState)
         .environmentObject(SettingsManager())
+        .environmentObject(RecordingLiveMetrics())
+        .environmentObject(previewAppState.liveStreamingText)
```
ray-amjad, line 1249: Minor (preview-only): this `#Preview` now injects `RecordingLiveMetrics` and `StreamingLiveText`, but `RecordingDialog` also declares `@EnvironmentObject var transcriptionPipeline: TranscriptionPipeline` (line 48), which is still not provided here. SwiftUI traps at render time with `No ObservableObject of type TranscriptionPipeline found` — so the preview still crashes even after this fix. It was already broken before this PR, but since the hunk is specifically about completing the environment for the dialog, worth closing the last gap.

```suggestion
        .environmentObject(RecordingLiveMetrics())
        .environmentObject(TranscriptionPipeline())
        .environmentObject(previewAppState.liveStreamingText)
```
(adjust the `TranscriptionPipeline` initializer to whatever its real init signature is)

ray-amjad:
Review pass (concurrency/correctness focus): **no correctness issues found.** Verified:

- All 13 `streamingText` references in `app/macos` are accounted for. Only `RecordingDialog` and `StreamingPreviewBubble` read it, and both were converted to observe `StreamingLiveText` directly — no view silently stops updating. All 9 write sites still mutate the same object through the `AppState.streamingText` passthrough.
- Both `RecordingDialog` construction sites (`RecordingWindowManager.swift:100`, `MainAppView.swift:280-282`) and the single `StreamingPreviewBubble` site (`StreamingPreviewWindowManager.swift:36`) inject `appState.liveStreamingText`, so no missing-`@EnvironmentObject` trap. `MainAppView` was in fact missing `audioManager.liveMetrics` on `main` — that would have crashed the dialog opened from the main window, and this PR incidentally fixes it.
- `liveStreamingText` is a `let`, so identity is stable across the panels' single-shot window construction; no stale-object capture.
- `AppState` is already `@MainActor`, so marking `StreamingLiveText` `@MainActor` adds no new isolation boundary — every write site was already main-actor-hopped before this PR.
- The new file needs no `project.pbxproj` entry; the Xcode project uses `PBXFileSystemSynchronizedRootGroup`.

One inline comment left on the `#Preview` block (minor, preview-only).

Pre-existing, **not** introduced here, so not blocking: the delta handlers in `RecordingTranscriptionFlow+Streaming.swift:463-507` dispatch each partial via an unstructured `Task { @MainActor in ... }`, which has no ordering guarantee — two deltas emitted close together can be applied out of order and briefly show a stale preview. Identical before and after this change; may be worth its own ticket.

ray-amjad:

---

### `app/macos/hyperwhisper/Views/RecordingDialog.swift:1340-1348` — ✅ **CONFIRMED**

**Original claim:** The #Preview injects RecordingLiveMetrics and StreamingLiveText but still omits TranscriptionPipeline, which RecordingDialog also declares as an @EnvironmentObject, so the preview traps at render even after this fix.

**What the code actually does:** RecordingDialog.swift:48 declares `@EnvironmentObject var transcriptionPipeline: TranscriptionPipeline` (alongside audioManager:45, appState:46, settingsManager:47, liveMetrics:51, liveStreamingText:53). The #Preview at RecordingDialog.swift:1340-1348 injects exactly five objects — AudioRecordingManager(), previewAppState, SettingsManager(), RecordingLiveMetrics(), previewAppState.liveStreamingText — and no TranscriptionPipeline. That property IS read at runtime: RecordingDialog.swift:329 does `transcriptActionHandler = TranscriptActionHandler(transcriptionPipeline: transcriptionPipeline)` inside the `.onAppear` block starting at line 302, so the preview canvas hits the `@EnvironmentObject` fatalError the moment the view appears. Severity caveat (verified, not a refutation): this is preview-only and pre-existing — `git show origin/main:.../RecordingDialog.swift` ends with a #Preview injecting only AudioRecordingManager/AppState/SettingsManager, i.e. main omits BOTH TranscriptionPipeline and RecordingLiveMetrics, so the PR strictly improves the preview rather than breaking it. It compiles (CI `Build & Test (macOS, unsigned)` is green on head ca2fe01) and has zero shipped-app impact, since the two real injection sites (RecordingWindowManager.swift:93-101 and MainAppView.swift:274-283) both pass transcriptionPipeline. One precision nit: the trap fires on `.onAppear`, one tick after first render, not literally during body evaluation — practically indistinguishable.


### `OVERALL` — 🟡 **PARTLY TRUE**

**Original claim:** PR is concurrency-clean and ships as-is

**What the code actually does:** Concurrency half CONFIRMED. (a) Isolation: AppState.swift:155-156 is `@MainActor class AppState: ObservableObject`; StreamingLiveText.swift:44-45 is `@MainActor final class StreamingLiveText: ObservableObject` with a single `@Published var text: String = ""` (line 50). Both sides of the passthrough at AppState.swift:263-266 (`get { liveStreamingText.text } / set { liveStreamingText.text = newValue }`) and the stored `let liveStreamingText = StreamingLiveText()` at AppState.swift:256 are on the same actor, so no new hop or isolation violation. (b) Reference sweep: `git grep streamingText origin/main -- app/macos` returns 21 hits, of which 9 are reactive view reads/writes (RecordingDialog 110/322/351/433, StreamingPreviewBubble 23/24/48/65/103) — all 9 are rewritten to `liveStreamingText.text` on this branch. The remaining 10 non-view hits (RecordingTranscriptionFlow+Streaming.swift 74/463/489/500/507/562/659/799, +StopRecording.swift:116, TranscriptionPipeline+PostProcessing.swift:58) are all pure WRITES via `appState?.streamingText = ...`, already wrapped in `Task { @MainActor in }` at the per-delta sites (Streaming.swift:462, 488, 498, 504) — none depend on AppState.objectWillChange, so nothing silently stops updating. No `$streamingText` publisher use anywhere. (c) Per-delta writes touch ONLY streamingText plus non-@Published locals (`streamingAccumulatedText`, `streamingPreviewTextSnapshot`, declared at RecordingTranscriptionFlow.swift:140), so the fix genuinely stops AppState from republishing per delta — the stated root cause is real and addressed. (d) Injection sites: `grep -rn 'RecordingDialog(\|StreamingPreviewBubble('` finds exactly three construction points — RecordingWindowManager.swift:93, MainAppView.swift:274, StreamingPreviewWindowManager.swift:34 — all three inject `appState.liveStreamingText`. (e) New file needs no pbxproj edit: project.pbxproj:80-90 uses PBXFileSystemSynchronizedRootGroup. CI check-runs on ca2fe01: 'Build & Test (macOS, unsigned): success'.

'Ships as-is' half REFUTED. `gh pr view 45 --json mergeable` returns CONFLICTING, and a real `git merge origin/main` produces 'CONFLICT (content): Merge conflict in app/macos/hyperwhisper/Views/MainAppView.swift'. Cause: main commit 0fec986 ('chore: remove dead code') deleted `openRecordingDialogWindow()` entirely — `git show origin/main:app/macos/hyperwhisper/Views/MainAppView.swift | grep 'openRecordingDialogWindow\|RecordingDialog('` returns nothing — so one of the PR's three injection hunks edits code that no longer exists. Also: the reviewer's '13 streamingText references' count is wrong; the base branch has 21.

**Corrected statement:** The PR is concurrency-clean — @MainActor isolation is consistent on both classes, all 21 (not 13) base-branch streamingText references are accounted for, all three RecordingDialog/StreamingPreviewBubble construction sites inject liveStreamingText, and macOS CI is green on head ca2fe01. But it does NOT ship as-is: it is CONFLICTING against main because commit 0fec986 deleted the `openRecordingDialogWindow()` function that the MainAppView.swift hunk modifies. Correct resolution is to drop that hunk (it was already dead code) and keep the RecordingWindowManager + StreamingPreviewWindowManager injections, which are the only two live paths.


---

PR #46, opened 2026-07-17
Scale slow-transcription thresholds to audio duration (HYPERWHISPER-P3/P4)

ray-amjad:
## Summary

Fixes noise-level Sentry perf warnings [HYPERWHISPER-P3](https://ray-amjad.sentry.io/issues/HYPERWHISPER-P3) ("Slow transcription completed") and [HYPERWHISPER-P4](https://ray-amjad.sentry.io/issues/HYPERWHISPER-P4) ("Slow transcribing UI completed").

**Verification first:** pulled a sample event — `transcribe=8366ms` of `totalMs=8456` (99% of time in the cloud transcribe call), but `silenceRemovedSeconds=161` shows this was a 3+ minute dictation (VAD trimmed 2.7 min of silence), not a short clip. P3 and P4 fire from the same underlying event (identical trace_id, same user, ~1s apart) — they're the same measurement surfaced at two points in the flow (pipeline transcribe stage vs. UI-completion layer), not two separate bugs.

Root cause: both thresholds (`TranscriptionPipeline+Transcription.swift` for P3, `RecordingTranscriptionFlow+StopRecording.swift` for P4) are flat cutoffs — 8s / 15s (post-processed) / 45s (local LLM) — regardless of audio length. ~8s to transcribe several minutes of audio via a cloud API is fine; the alarm was on absolute time, not throughput.

## Fix

- Add a per-audio-second allowance (`150ms/s`) on top of each base threshold, so the bar scales with the audio actually sent to the provider (VAD-trimmed length when VAD ran, else raw recording) instead of penalizing long-but-proportionally-fast calls.
- Thread the resolved duration from the recording flow into `transcribeWithDetails(audioDurationSeconds:)` so P3 and P4 use the same value; other callers (retry, file-import) fall back to reading the file's duration directly, so they get the same fix for free.
- Surface `audioDurationSeconds` / `durationAllowanceMs` / effective threshold in the Sentry extras attached to any slow-path event, so a real regression (not just a long recording) is easier to triage going forward.

No upload-compression change included — the cloud path already streams raw binary (no multipart overhead per the existing comment in `HyperWhisperCloudProvider`), and 8s for ~2 minutes of trimmed audio is a reasonable ratio, so that wasn't the bottleneck.

## Test plan

- [ ] Could not run `xcodebuild` in this sandbox (Linux, no Xcode toolchain) — needs a macOS build/typecheck pass before merge.
- [ ] Manually dictate a short (~5s) and a long (~3min) recording and confirm neither trips the warning log line under normal cloud latency.
- [ ] Confirm a genuinely slow cloud response (e.g. throttled network) still trips the threshold for both short and long recordings.

### DIFF
--- a/app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
+++ b/app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
@@ -391,7 +391,7 @@ extension RecordingTranscriptionFlow {
             // Audio actually sent to the provider: VAD-trimmed length when VAD ran,
             // else the raw recording. Reused below for the UI-side slow threshold so
             // both P3 (pipeline) and P4 (this flow) scale off the same duration.
-            let effectiveAudioDurationSeconds = trimResult?.trimmedDuration ?? recordingDuration
+            let effectiveAudioDurationSeconds = (vadResult.wasProcessed ? trimResult?.trimmedDuration : nil) ?? recordingDuration
             let transcribeStart = Date()
             let transcriptionResult = try await transcriptionMgr.transcribeWithDetails(
                 audioURL: finalAudioURL,
--- a/app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
+++ b/app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
@@ -53,9 +53,14 @@ extension RecordingTranscriptionFlow {
         var createRowMs = -1
         var transcribeMs = -1
         var coreDataUpdateMs = -1
+        // HYPERWHISPER-P4: same fix as HYPERWHISPER-P3 (TranscriptionPipeline+Transcription.swift) —
+        // these base thresholds cover fixed overhead; `slowTranscribingUIPerAudioSecondMs` adds a
+        // per-audio-second allowance so long-but-proportionally-fast dictations stop tripping the
+        // same bar as short ones.
         let slowTranscribingUIThresholdMs = 8_000
         let slowTranscribingUIWithPostProcessingThresholdMs = 15_000
         let slowTranscribingUIWithLocalLLMThresholdMs = 45_000
+        let slowTranscribingUIPerAudioSecondMs = 150.0
         var transcribingUIStart: Date?
 
         let sessionModeName = activeSessionModeName
@@ -383,12 +388,17 @@ extension RecordingTranscriptionFlow {
             // Use finalAudioURL which may be VAD-trimmed if VAD was enabled.
             // `transcribe` folds provider selection + transcription + post-processing;
             // its wall time is surfaced as the transcribe stage.
+            // Audio actually sent to the provider: VAD-trimmed length when VAD ran,
+            // else the raw recording. Reused below for the UI-side slow threshold so
+            // both P3 (pipeline) and P4 (this flow) scale off the same duration.
+            let effectiveAudioDurationSeconds = trimResult?.trimmedDuration ?? recordingDuration
             let transcribeStart = Date()
             let transcriptionResult = try await transcriptionMgr.transcribeWithDetails(
                 audioURL: finalAudioURL,
                 mode: transcriptionMode,
                 recordingSession: nil, // Will be set by RecordingLifecycle
-                applicationContext: capturedApplicationContext
+                applicationContext: capturedApplicationContext,
+                audioDurationSeconds: effectiveAudioDurationSeconds
             )
             transcribeMs = Int(Date().timeIntervalSince(transcribeStart) * 1000)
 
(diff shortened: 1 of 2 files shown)


app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift
```diff
@@ -383,12 +388,17 @@ extension RecordingTranscriptionFlow {
             // Use finalAudioURL which may be VAD-trimmed if VAD was enabled.
             // `transcribe` folds provider selection + transcription + post-processing;
             // its wall time is surfaced as the transcribe stage.
+            // Audio actually sent to the provider: VAD-trimmed length when VAD ran,
+            // else the raw recording. Reused below for the UI-side slow threshold so
+            // both P3 (pipeline) and P4 (this flow) scale off the same duration.
+            let effectiveAudioDurationSeconds = trimResult?.trimmedDuration ?? recordingDuration
```
ray-amjad: **`trimResult` is non-nil even when the trimmed audio was rejected, so this can pick a duration for a file that was never sent.**

`VADProcessingService` returns the `TrimResult` on the validation-failure path too — it only zeroes `wasProcessed` and keeps `finalAudioURL` as the *original* URL:

```swift
} else {
    // Validation failed, use original audio
    return VADProcessingResult(
        finalAudioURL: audioURL,
        trimResult: trimResult,   // <-- still non-nil
        wasProcessed: false,
        originalAudioURL: audioURL
    )
}
```
(`VADProcessingService.swift:228-234`)

So `trimResult?.trimmedDuration ?? recordingDuration` uses the trimmed length while `finalAudioURL` is the full untrimmed recording.

Concrete case: VAD is on, user records 40s that is mostly silence/noise, `SilenceTrimmer` yields `trimmedDuration = 0.2s`, which fails `validateTrimResult` check 2 (`>= minimumTrimmedDuration`). The full 40s file is sent to the provider, but `effectiveAudioDurationSeconds = 0.2`, so the allowance is 30ms instead of 6000ms — the exact long-recording false positive this PR is fixing still fires, and the new `stage_audio_duration_seconds` / `audioDurationSeconds` Sentry extras report 0.2s for a 40s recording, which is actively misleading during triage. Check 3 (trimmed file missing/undersized) has the same shape.

Note the existing code two lines up already guards on `wasProcessed` for exactly this reason:
`trimmedAudioPath: (vadResult.wasProcessed ? trimResult?.outputURL.path : nil)` (line 337).

```suggestion
            let effectiveAudioDurationSeconds = (vadResult.wasProcessed ? trimResult?.trimmedDuration : nil) ?? recordingDuration
```

ray-amjad:

**Result: 1 confirmed.**

---

### `app/macos/hyperwhisper/Managers/AudioRecording/RecordingFlow/RecordingTranscriptionFlow+StopRecording.swift:394` — ✅ **CONFIRMED**

**Original claim:** trimResult is non-nil even when the VAD trim was REJECTED (wasProcessed=false, finalAudioURL reverts to the original file), so the slow-transcription threshold can be computed from a duration belonging to a file that was never sent.

**What the code actually does:** VADProcessingService.swift:229-237 is the validation-failure branch: it returns `VADProcessingResult(finalAudioURL: audioURL /* the ORIGINAL */, trimResult: trimResult /* NON-NIL */, wasProcessed: false, originalAudioURL: audioURL)`. So a non-nil trimResult does NOT imply the trimmed file was used. (Contrast the two genuinely-nil paths: :239-250 `catch TrimError.noSpeechDetected` and :252-277 the general catch, both of which pass `trimResult: nil`; and :179-189, the VAD-disabled/too-short guard, also nil.) `validateTrimResult` (:295-374) returns nil — driving that branch — in three cases: silenceRemoved <= AudioConstants.minimumSilenceRemoved (0.5s, AudioConstants.swift:70), trimmedDuration < AudioConstants.minimumTrimmedDuration (0.3s, AudioConstants.swift:79), or the trimmed file missing/<5KB (AudioConstants.swift:62). In RecordingTranscriptionFlow+StopRecording.swift, `let trimResult = vadResult.trimResult` at :325 is bound once and never reassigned (grep confirms :325/:337/:394/:560 are the only references), and :394 reads `trimResult?.trimmedDuration ?? recordingDuration` with no `vadResult.wasProcessed` gate — even though the added comment at :391-393 states the intent is 'audio actually sent to the provider'. The decisive corroboration is 57 lines earlier in the SAME file: :337 writes `trimmedAudioPath: (vadResult.wasProcessed ? trimResult?.outputURL.path : nil)` — the existing code explicitly gates on `wasProcessed` before trusting the same optional, precisely because non-nil-ness is not sufficient. The wrong duration then propagates further: it is passed as `audioDurationSeconds:` into `transcribeWithDetails` at :401, and TranscriptionPipeline+Transcription.swift:88-89 short-circuits its own AVURLAsset measurement whenever that value is > 0, so the stale trimmed duration silently drives the P3 pipeline threshold too. Worst realistic case: a 120s recording whose trim is rejected by the <0.3s check yields an allowance of Int(0.29 * 150) = 43ms instead of Int(120 * 150) = 18000ms, so effectiveUIThreshold at :575 is ~18s too tight and fires the very spurious 'slow' Sentry event this PR exists to suppress. Severity is telemetry-only (no user-visible or transcription-correctness impact) and the common rejection case (silenceRemoved <= 0.5s) skews the allowance by at most 75ms, but the mechanism and the stated consequence are both real. Fix: `let effectiveAudioDurationSeconds = vadResult.wasProcessed ? (trimResult?.trimmedDuration ?? recordingDuration) : recordingDuration`.


---

PR #48, opened 2026-07-17
Fix Parakeet fallback contract when model isn't downloaded (HYPERWHISPER-TD, HYPERWHISPER-SV)

ray-amjad:
## Problem

Sentry HYPERWHISPER-TD (\"Parakeet prepare failed\", 121 events / 28 users, still firing on 2.41.0) and HYPERWHISPER-SV (\"model_not_downloaded at select_provider\") share a root cause: a Parakeet-backed transcription mode gets selected/prepared while no Parakeet model is downloaded locally (or it was evicted from disk), and the code logs \"will use cloud fallback\" but then still calls straight into Parakeet's \`prepareIfNeeded\`, which throws \`modelNotDownloaded\` instead of actually falling back to cloud.

Two call sites had this bug:
- \`TranscriptionModelManager.prepareModel\` (Parakeet branch, ~line 196-204) — invoked from app launch / mode switch / record-start prewarm. The generic catch here is what reports \"Failed to prepare Parakeet provider\" to Sentry as HYPERWHISPER-TD.
- \`TranscriptionProviderRouter.selectLocalProvider\` (~line 711-725) — the actual \`select_provider\` transcription-time path, root cause of HYPERWHISPER-SV.

## Fix

- \`TranscriptionProviderRouter.selectLocalProvider\`: check \`parakeetModelManager.isModelInstalled(modelId)\` before touching the runtime. If not installed, kick off \`startDownload(modelId)\` (same API the Models UI already uses) and return a real cloud fallback provider instead of throwing. Added a defensive catch for \`modelNotDownloaded\` around the actual \`prepareIfNeeded\` call to cover a disk-eviction race between the check and the call.
- \`TranscriptionModelManager.prepareModel\`: added a specific catch for \`modelNotDownloaded\` in the Parakeet branch that starts the download, sets ready state to a real \"Cloud\" fallback, and surfaces a plain-language status message — instead of falling into the generic catch that Sentry-reports it as a crash.

Net effect: the \"will use cloud fallback\" log message is now backed by real behavior, and selecting a Parakeet mode with no local model auto-triggers a download instead of silently failing.

## Verification

No macOS/Xcode toolchain available in this sandbox to compile/run tests. Traced types manually end-to-end (including fixing an ambiguous-overload issue on \`ensureHealthy(.hyperwhisper)\` introduced by the cloud-fallback path). No existing test harness covers this MainActor-isolated, FluidAudio-dependent routing code (existing tests only cover pure logic elsewhere), so no test was added rather than ship unverifiable scaffolding. **Recommend a real Xcode build/typecheck on this branch before merging.**

Refs: HYPERWHISPER-TD, HYPERWHISPER-SV

---

### DIFF
--- a/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
@@ -193,6 +193,22 @@ class TranscriptionModelManager: ObservableObject {
                 modelReadyState = .ready(name: parakeetDisplayName)
             } catch is CancellationError {
                 AppLogger.models.info("Parakeet preparation cancelled")
+            } catch TranscriptionError.modelNotDownloaded {
+                // EXPECTED, NOT A CRASH (HYPERWHISPER-TD): the selected mode
+                // points at a Parakeet model that isn't on disk (fresh
+                // install, or evicted after being downloaded). This used to
+                // fall into the generic `catch` below, which Sentry-reported
+                // it as "Parakeet prepare failed" on every launch/mode-switch
+                // without actually doing anything about it — the "will use
+                // cloud fallback" log never turned into real fallback
+                // behavior. Kick off the same download the Models UI uses,
+                // tell the user their local mode needs a download, and mark
+                // the model ready as Cloud so transcription isn't blocked
+                // while the download completes in the background.
+                AppLogger.models.warning("Parakeet \(modelId, privacy: .public) not downloaded — starting download, using cloud fallback in the meantime")
+                parakeetModelManager.startDownload(modelId)
+                modelReadyState = .ready(name: "Cloud")
+                onStateChange?(.error(message: "\(parakeetDisplayName) isn't downloaded yet — downloading now. Using cloud until it's ready."))
             } catch {
                 AppLogger.models.error("Failed to prepare Parakeet provider: \(error.localizedDescription, privacy: .public)")
                 modelReadyState = .none
--- a/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
@@ -69,6 +69,7 @@ class TranscriptionModelManager: ObservableObject {
         case none
         case loading(name: String)
         case ready(name: String)
+        case unavailable(name: String)
     }
 
     // MARK: - Published Properties
@@ -194,21 +195,14 @@ class TranscriptionModelManager: ObservableObject {
             } catch is CancellationError {
                 AppLogger.models.info("Parakeet preparation cancelled")
             } catch TranscriptionError.modelNotDownloaded {
-                // EXPECTED, NOT A CRASH (HYPERWHISPER-TD): the selected mode
-                // points at a Parakeet model that isn't on disk (fresh
-                // install, or evicted after being downloaded). This used to
-                // fall into the generic `catch` below, which Sentry-reported
-                // it as "Parakeet prepare failed" on every launch/mode-switch
-                // without actually doing anything about it — the "will use
-                // cloud fallback" log never turned into real fallback
-                // behavior. Kick off the same download the Models UI uses,
-                // tell the user their local mode needs a download, and mark
-                // the model ready as Cloud so transcription isn't blocked
-                // while the download completes in the background.
-                AppLogger.models.warning("Parakeet \(modelId, privacy: .public) not downloaded — starting download, using cloud fallback in the meantime")
-                parakeetModelManager.startDownload(modelId)
-                modelReadyState = .ready(name: "Cloud")
-                onStateChange?(.error(message: "\(parakeetDisplayName) isn't downloaded yet — downloading now. Using cloud until it's ready."))
+                // Missing weights are expected for a restored mode or an
+                // externally-evicted cache. Keep the local-only contract:
+                // preparation must neither upload audio nor start a large
+                // download without consent. The status bar makes the required
+                // action visible, while an actual transcription attempt
+                // throws the localized `.modelNotDownloaded` error.
+                AppLogger.models.warning("Parakeet \(modelId, privacy: .public) is not downloaded")
+                modelReadyState = .unavailable(name: parakeetDisplayName)
             } catch {
                 AppLogger.models.error("Failed to prepare Parakeet provider: \(error.localizedDescription, privacy: .public)")
                 modelReadyState = .none
--- a/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
@@ -447,7 +447,14 @@ class TranscriptionProviderRouter {
             }
             modelString = m
         case "parakeet":
-            modelString = model?.isEmpty == false ? model! : "parakeet-tdt-v3-multilingual"
+            let requestedModelId = ParakeetModelManager.Constants.modelIdForSelection(model)
+            guard let canonicalModelId = ParakeetModelManager.Constants.canonicalModelId(for: requestedModelId) else {
+                throw TranscriptionError.providerNotAvailable(
+                    provider: "Parakeet",
+                    reason: "Unknown Parakeet model '\(requestedModelId)'"
+                )
+            }
+            modelString = canonicalModelId
         case "qwen3asr", "qwen3", "qwen3-asr":
             modelString = Qwen3AsrModelManager.Constants.modelId
         case "applespeech", "apple", "apple-speech", "apple-speech-analyzer", "speech-analyzer":
@@ -645,38 +652,6 @@ class TranscriptionProviderRouter {
         modelId == "apple-speech-analyzer"
     }
 
-    /// Safety-net fallback used when a local model is selected for this
-    /// transcription but isn't actually downloaded (e.g. a mode still points
-    /// at a Parakeet model that was evicted from disk). Always resolves to
-    /// the built-in HyperWhisper Cloud provider — the same "no local model,
-    /// cloud fallback" promise the Models scan log makes — rather than
-    /// whatever BYOK cloud provider the mode's `cloudProvider` field might
-    /// name, since that field only applies when `mode.model == "cloud"`.
-    private func selectCloudFallbackProvider() async throws -> TranscriptionProvider {
-        ensureHyperWhisperCloudProvider()
-        guard let hwProvider = hyperwhisperCloudProvider else {
-            throw TranscriptionError.providerNotAvailable(provider: "HyperWhisper Cloud", reason: "Failed to initialize provider")
-        }
-
-        if let healthManager = providerHealthManager {
-            let status = await healthManager.ensureHealthy(CloudProvider.hyperwhisper)
-            if !status.isHealthy {
-                AppLogger.network.error("Cloud fallback provider not healthy · status=\(status.statusText, privacy: .public)")
-                throw errorForHealthStatus(status, providerDisplayName: CloudProvider.hyperwhisper.displayName)
-            }
-        }
-
-        guard hwProvider.isAvailable else {
-            throw TranscriptionError.providerNotAvailable(provider: hwProvider.name, reason: "Provider is not configured or ready")
-        }
-
-        guard NetworkStatus.shared.isOnline else {
-            throw TranscriptionError.transientNetwork(details: "No internet connection")
-        }
-
-        return hwProvider
-    }
-
     /// Select local provider based on model ID
     /// DECISION TREE:
     /// - If model ID maps to WhisperModel enum → Use LibWhisperProvider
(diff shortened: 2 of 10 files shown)


app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
```diff
@@ -645,6 +645,38 @@ class TranscriptionProviderRouter {
...
+
+        guard hwProvider.isAvailable else {
+            throw TranscriptionError.providerNotAvailable(provider: hwProvider.name, reason: "Provider is not configured or ready")
+        }
+
+        guard NetworkStatus.shared.isOnline else {
+            throw TranscriptionError.transientNetwork(details: "No internet connection")
+        }
+
+        return hwProvider
```
ray-amjad: **Uploads audio to the cloud for a user who never opted into cloud, with zero entitlement check.**

None of the three guards here actually gate anything for HyperWhisper Cloud:
- `HyperWhisperCloudProvider.isAvailable` is hardcoded `{ true }` (`Providers/Cloud/HyperWhisperCloudProvider.swift:67`).
- `CloudProviderHealthManager.performHealthCheck` short-circuits `.hyperwhisper` to `.healthy` unconditionally (`Managers/CloudProviderHealthManager.swift:469-471`).
- `HyperWhisperCloudProvider.transcribe` uploads with a **device ID** when no license exists (`getTranscriptionIdentifier()` returns `(identifier, isLicensed)` and proceeds either way, line 224).

So the effective policy becomes: "any recording in a Parakeet mode whose model is missing is uploaded to the Fly backend, licensed or not."

Concrete scenario: a privacy-motivated user picks Parakeet V3 precisely so audio never leaves the machine, then deletes/never downloads the weights (or the cache is evicted). They press record, get no prompt, and the raw audio is POSTed to HyperWhisper Cloud. A licensed user silently burns metered credits on a mode they configured as free/local — note `RecordingTranscriptionFlow+StartRecording.swift:359` explicitly documents "No local license/trial gate — local transcription is unlimited", a gate this path now routes around. An unlicensed user has the audio uploaded and *then* gets a server-side credit/auth rejection, so the privacy cost is paid for a transcription that fails anyway.

Also note every other local engine (LibWhisper `LibWhisperProvider.swift:115`, Qwen3, Nemotron, Apple Speech) still hard-fails with `.modelNotDownloaded` in this situation, so Parakeet becomes the one engine that silently goes off-device.

Suggest gating this on an explicit, user-visible opt-in (a settings flag and/or a one-time consent prompt) and on a real entitlement check, rather than falling back implicitly.

app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
```diff
@@ -718,10 +750,39 @@ class TranscriptionProviderRouter {
...
+            // Previously this fell straight into `prepareIfNeeded`, which
+            // throws `.modelNotDownloaded` and fails the whole transcription
+            // instead of honouring the "no local model, will use cloud
+            // fallback" contract. Kick off the download (same `startDownload`
+            // the Models UI uses) and route *this* transcription to
+            // HyperWhisper Cloud so the user still gets a result.
+            if let parakeetModelManager, !parakeetModelManager.isModelInstalled(modelId) {
+                AppLogger.transcription.warning("⚠️ Parakeet model not downloaded (\(modelId, privacy: .public)) — starting download and falling back to cloud for this transcription")
+                await MainActor.run {
+                    parakeetModelManager.startDownload(modelId)
```
ray-amjad: **Auto-`startDownload` re-downloads a model the user deliberately deleted, with no way to refuse.**

`ParakeetModelManager.deleteModel(_:)` only removes the cache directory — it does not rewrite `mode.model`. So the sequence is:

1. User opens Models and deletes Parakeet V3 to reclaim disk (multi-GB bundle).
2. Any of: app launch (`hyperwhisperApp.swift:745`), a mode switch (`AppState.swift:844`), or simply pressing the record hotkey (record-start prewarm, `RecordingTranscriptionFlow+StartRecording.swift:352`) calls `prepareModel`, or a transcription hits this line.
3. `startDownload(modelId)` fires and the multi-GB download begins again immediately, unattended, on whatever network they are on (tethered/metered included).

The only escape is editing every mode that references the model. That inverts the meaning of the Models delete button, and it fires from a hotkey path where the user never gets a chance to decline. `DownloadController.start` de-dupes concurrent starts so it will not thrash within a session, but it restarts after every cancel.

Recommend an actionable toast ("Parakeet V3 isn't downloaded — Download") rather than initiating a large transfer implicitly.

app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
```diff
@@ -718,10 +750,39 @@ class TranscriptionProviderRouter {
...
+                }
+                return try await selectCloudFallbackProvider()
+            }
+
+            do {
+                // Pass specific modelId to prepare the correct version (V2 or V3)
+                try await parakeetProvider.prepareIfNeeded(language: language, modelId: modelId)
+                AppLogger.transcription.info("✅ Parakeet provider selected for model: \(modelId)")
+                return parakeetProvider
+            } catch TranscriptionError.modelNotDownloaded {
```
ray-amjad: **The defensive catch does not actually cover the race it describes.**

`ParakeetProvider.prepareIfNeeded` only throws `.modelNotDownloaded` from its up-front `AsrModels.modelsExist(at:)` guard (`Providers/Local/ParakeetProvider.swift:212-217`). Everything after that — the `runtime.currentManager(for:)` load — is rewritten:

```swift
} catch {
    await runtime.reset()
    throw TranscriptionError.providerNotAvailable(provider: "Parakeet", reason: "Failed to initialize Parakeet runtime")
}
```

So if the weights are evicted (or are still being written by the download this PR now kicks off in the background) *after* the `modelsExist` check and the CoreML load then fails, the thrown error is `.providerNotAvailable`, not `.modelNotDownloaded` — this `catch` does not match and the transcription hard-fails, which is the exact failure mode the PR is trying to eliminate. The window between the `isModelInstalled` gate at line 763 and `modelsExist` inside `prepareIfNeeded` is microseconds; the load window is seconds, so the uncovered branch is the likelier one.

Worth also matching `.providerNotAvailable(provider: "Parakeet", _)` here (or having `prepareIfNeeded` preserve the underlying error kind rather than flattening it).

app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
```diff
@@ -193,6 +193,22 @@ class TranscriptionModelManager: ObservableObject {
...
+                // without actually doing anything about it — the "will use
+                // cloud fallback" log never turned into real fallback
+                // behavior. Kick off the same download the Models UI uses,
+                // tell the user their local mode needs a download, and mark
+                // the model ready as Cloud so transcription isn't blocked
+                // while the download completes in the background.
+                AppLogger.models.warning("Parakeet \(modelId, privacy: .public) not downloaded — starting download, using cloud fallback in the meantime")
+                parakeetModelManager.startDownload(modelId)
+                modelReadyState = .ready(name: "Cloud")
+                onStateChange?(.error(message: "\(parakeetDisplayName) isn't downloaded yet — downloading now. Using cloud until it's ready."))
```
ray-amjad: **This user-facing message is never displayed — nothing observes `TranscriptionPipeline.state`.**

`onStateChange` is wired to `self?.state = newState` (`TranscriptionPipeline.swift:302-304`), and `state` is only ever read back by `errorMessage`/`transcriptionProgress`/`state_isReadyForTranscription()`. No view observes it (the status UI observes `modelCoordinator.$modelReadyState` — `Views/Components/ModelStatusBar.swift:56` — and toasts come from `AppState.showError` on the `RecordingState` path, which this never touches).

So the stated goal of the comment above ("tell the user their local mode needs a download") does not happen: the only thing the user sees is the model chip flipping to "Cloud", with no explanation that a multi-GB download just started or that this recording is going off-device. That silence is what makes the auto-download and the cloud routing invisible.

Two further consequences of using `.error` for an informational message:
- `prepareModel` runs from the record-start prewarm (`RecordingTranscriptionFlow+StartRecording.swift:352`), so the pipeline is pushed into `.error` while a recording is in flight. It survives today only because `state_isReadyForTranscription()` treats `.error` as ready (`TranscriptionPipeline+Transcription.swift:446-452`); it is a fragile invariant to lean on for a non-error.
- If someone later wires a toast to `errorMessage` (the obvious fix for the point above), this fires on every launch, every mode switch and every record-start for the whole duration of the download.

Route it through `AppState.showError`/an informational toast instead, and keep `.error` for real failures.

ray-amjad:
**Overlaps PR #52 (`fix/sv-td-parakeet-cloud-fallback`) — same two files, same two Sentry issues, mutually exclusive implementations.**

#52 patches the identical two call sites with a `guard parakeetProvider.isAvailable(for: modelId)` and routes to `hyperwhisperCloudProvider`, but it does **not** auto-start a download and does **not** run the health/`isAvailable`/network pre-checks this PR adds. This PR is a strict superset behaviourally, and the two diffs touch overlapping lines in both files, so they will conflict textually — merging both would leave two stacked availability gates in `selectLocalProvider` and two "not downloaded" branches in `prepareModel`.

Pick one. If this one wins, the auto-`startDownload` and the implicit cloud routing are the parts that still need a decision (see the inline comments): #52 deliberately does neither, which side-steps the "re-download a model the user deleted" and "upload audio for a local-only mode" questions, at the cost of leaving the user stuck on cloud with no path back to local.

ray-amjad:

**Result: 2 partly true, 1 refuted.**

---

### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift:677` — 🟡 **PARTLY TRUE**

**Original claim:** selectCloudFallbackProvider uploads audio to HyperWhisper Cloud for a user who never opted into cloud with ZERO entitlement check — isAvailable hardcoded true, health check short-circuits .hyperwhisper to healthy, transcribe falls back to a device ID; violates the server-side-entitlement rule.

**What the code actually does:** All three sub-claims are literally true. (1) HyperWhisperCloudProvider.swift:67 — `var isAvailable: Bool { true }`, so the guard at router:669 can never fail. (2) CloudProviderHealthManager.swift:469-471 — `if provider == .hyperwhisper || .microsoftAzureSpeech || .googleSpeech { return .healthy }` inside performHealthCheck, so ensureHealthy(.hyperwhisper) at router:662 is always healthy. (3) HyperWhisperCloudProvider.swift:224 calls `licenseManager.getTranscriptionIdentifier()` and lines 306-307 pass `licenseKey: isLicensed ? identifier : nil, deviceID: isLicensed ? nil : identifier` (LicenseManager.swift:191-198 returns DeviceIdentifierGenerator.generate() when unlicensed). BUT the entitlement framing is refuted: hyperwhisper-cloud/src/middleware/auth.ts:99-104 rejects any request without a license key (`licenseRequiredResponse()`) and routes/transcribe.ts:371 runs validateAuth BEFORE reading the audio body at line 406 — so an unlicensed user gets a 401, never a free transcription; no client-side bypass is introduced. The fallback also applies exactly the same three client-side gates the user-chosen cloud path already applies (router:538-551: ensureHealthy → provider.isAvailable → NetworkStatus), so it is not a weakened path.

**Corrected statement:** The three cited client-side gates are indeed no-ops for HyperWhisper Cloud, but this is not an entitlement bypass — entitlement is enforced server-side (auth.ts:99-104), and the fallback reuses the identical gate sequence as the normal cloud path (router:538-551). The real defect is narrower: a mode the user configured as local-only silently ships its audio off-device with no opt-in, spending a licensed user's credits; an unlicensed user gets a confusing 'license required' error instead of 'Parakeet model not downloaded'.


### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift:766` — 🟡 **PARTLY TRUE**

**Original claim:** auto-startDownload re-downloads a multi-GB model the user deliberately deleted, triggerable by launch, mode switch, or the record hotkey, with no way to refuse.

**What the code actually does:** The unattended-download mechanism and all three triggers are real: router:763-768 and TranscriptionModelManager.swift:208-209 both call `parakeetModelManager.startDownload(modelId)` with no confirmation; prepareModel is invoked from hyperwhisperApp.swift:745 (launch), AppState.swift:844 ($selectedModeId sink = mode switch), and RecordingTranscriptionFlow+StartRecording.swift:352 (record-start prewarm). But the size and 'deliberately deleted / no way to refuse' parts are wrong: ParakeetModelManager.swift:47/55 declare the models as "474 MB" (V2) and "494 MB" (V3), not multi-GB; ModelLibraryView.swift:788-793 (checkAndRemoveParakeetModel → showCannotDeleteAlertIfNeeded) BLOCKS deleting a Parakeet model while any Mode points at it, and ModesView.swift:58-63 only offers already-downloaded models in the mode editor, so the in-app 'deliberately deleted while a mode still uses it' state is not reachable — the realistic trigger is off-disk eviction/manual Finder deletion or a restored backup; and a running download can be refused via ModelLibraryView.swift:709 → ParakeetModelManager.cancelDownload:203 → DownloadController.cancel:55. DownloadController.start:43 also guards `tasks[key] == nil` so repeated triggers don't stack downloads.

**Corrected statement:** Accurate statement: the PR starts an unprompted ~0.5 GB (474/494 MB) Parakeet download from three fire-and-forget paths (launch, mode switch, record-start prewarm) with no consent dialog, and it will restart on the next trigger after a user cancels. It is not multi-GB, it is not normally reachable by in-app deletion (which is blocked while a mode uses the model), and the user can cancel an in-flight download from the Model Library.


### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift:776` — ❌ **REFUTED**

**Original claim:** The defensive `catch TranscriptionError.modelNotDownloaded` does not cover the race it claims to; the runtime-load path rethrows .providerNotAvailable, not .modelNotDownloaded.

**What the code actually does:** ParakeetProvider.prepareIfNeeded re-checks the disk itself before loading: ParakeetProvider.swift:213-217 does `let directory = AsrModels.defaultCacheDirectory(for: targetVersion); guard AsrModels.modelsExist(at: directory) else { throw TranscriptionError.modelNotDownloaded }`. That guard is exactly the check-then-use window the comment describes (eviction between the router's isModelInstalled check at :763 and prepare at :773), and it throws .modelNotDownloaded, which the new catch at :776 does receive (TranscriptionError.swift:10-12 is a plain enum case, so the enum-case catch pattern matches). Only the much narrower window where files disappear AFTER that guard passes and during runtime load hits ParakeetProvider.swift:223-226, which maps the failure to .providerNotAvailable — and that case failed the transcription identically before this PR. Note also that the router's isModelInstalled (ParakeetModelManager.swift:131-133) reads cached `availableModels` state, whereas prepareIfNeeded stats the disk, so the catch additionally covers a stale-cache mismatch, not just a true race.

**Corrected statement:** The defensive catch does cover the described race: prepareIfNeeded's own on-disk `AsrModels.modelsExist` guard (ParakeetProvider.swift:214-216) throws .modelNotDownloaded for an eviction between the router's check and prepare. Only a sub-window during CoreML runtime load surfaces as .providerNotAvailable (ParakeetProvider.swift:226), which is pre-existing behaviour and not the race the comment claims to handle.


---

PR #49, opened 2026-07-17
Fix Windows NRE in backup export settings (HYPERWHISPER-TK)

ray-amjad:
## Problem
HYPERWHISPER-TK — NullReferenceException in `BackupExportSettingsPage.BuildExportSelection()` on Windows. 17 users, 31 events, last seen 2026-07-13.

## Root cause
Checkboxes in `BackupExportSettingsPage.xaml` (`ExportSettingsCheckbox`, `ExportModesCheckbox`, `ExportVocabularyCheckbox`) are declared with `IsChecked="True"`. Setting that in XAML fires `Checked` → `ExportSection_Changed` → `UpdateExportButtonState()` synchronously *during* `InitializeComponent()`, before sibling checkboxes further down the visual tree have been constructed. `BuildExportSelection()` then dereferences a checkbox that doesn't exist yet — a classic WPF construction-order bug, matching the reported stack (`BuildExportSelection` ← `UpdateExportButtonState` ← `ExportSection_Changed`, fired from the constructor).

The code already guarded on `IsInitialized`, but that relies on undocumented WPF timing (whether `FrameworkElement.IsInitialized` flips before or after nested BAML parsing completes) rather than an explicit contract.

## Fix
Replace the `IsInitialized` check with an explicit `_initialized` boolean field, set as the final statement in the constructor after `InitializeComponent()` and setup is complete. `UpdateExportButtonState()` now bails out until that flag is set, guaranteeing `BuildExportSelection()` never runs before the full control tree exists — independent of any WPF-internal timing assumption.

## Scope
UI/event-handler timing fix only — no backup field/schema changes, so `shared-backup/` was not touched (checked its CLAUDE.md; not applicable here).

## Verification
`dotnet build` on `HyperWhisper.csproj` succeeds with 0 errors (pre-existing native-DLL guard for an unrelated Rust component worked around only for the build check, not part of the diff).

---

### DIFF
--- a/app/windows/HyperWhisper.SmokeTests/Program.cs
+++ b/app/windows/HyperWhisper.SmokeTests/Program.cs
@@ -351,9 +351,39 @@ internal static class Program
                 var application = new Application();
                 LoadApplicationResources(application);
 
+                // Constructing the page exercises the exact construction-order NRE the
+                // `_initialized` guard fixes: IsChecked="True" on the export checkboxes
+                // fires ExportSection_Changed synchronously *during* InitializeComponent,
+                // before sibling checkboxes further down the visual tree exist. If the
+                // guard is ever removed, or `_initialized` is set before InitializeComponent
+                // finishes, BuildExportSelection dereferences a not-yet-created checkbox and
+                // this constructor call throws NullReferenceException.
                 var page = new BackupExportSettingsPage();
                 if (!page.IsInitialized)
                     throw new InvalidOperationException("BackupExportSettingsPage did not finish WPF initialization.");
+
+                // All three export checkboxes default to IsChecked="True" in XAML, so
+                // UpdateExportButtonState should already have computed HasAnySection == true
+                // and left the export button enabled. This only holds if `_initialized`
+                // was set correctly by the time the constructor finished.
+                Assert(page.ExportButton.IsEnabled,
+                    "expected ExportButton to be enabled after construction (all sections checked)");
+
+                // Exercise the guard again post-construction: unchecking every export
+                // section should disable the button, and re-checking one should re-enable
+                // it. This fails if `_initialized` is ever re-inverted (e.g.
+                // `if (_initialized) return;`) or never set, since UpdateExportButtonState
+                // would then never run and the button would stay stuck at its initial state
+                // regardless of checkbox changes.
+                page.ExportSettingsCheckbox.IsChecked = false;
+                page.ExportModesCheckbox.IsChecked = false;
+                page.ExportVocabularyCheckbox.IsChecked = false;
+                Assert(!page.ExportButton.IsEnabled,
+                    "expected ExportButton to be disabled once all export sections are unchecked");
+
+                page.ExportModesCheckbox.IsChecked = true;
+                Assert(page.ExportButton.IsEnabled,
+                    "expected ExportButton to be re-enabled after re-checking a section");
 
                 application.Shutdown();
             });
--- a/app/windows/HyperWhisper.SmokeTests/Program.cs
+++ b/app/windows/HyperWhisper.SmokeTests/Program.cs
@@ -351,30 +351,15 @@ internal static class Program
                 var application = new Application();
                 LoadApplicationResources(application);
 
-                // Constructing the page exercises the exact construction-order NRE the
-                // `_initialized` guard fixes: IsChecked="True" on the export checkboxes
-                // fires ExportSection_Changed synchronously *during* InitializeComponent,
-                // before sibling checkboxes further down the visual tree exist. If the
-                // guard is ever removed, or `_initialized` is set before InitializeComponent
-                // finishes, BuildExportSelection dereferences a not-yet-created checkbox and
-                // this constructor call throws NullReferenceException.
+                // Constructing the page exercises the exact construction-order NRE this
+                // regression test covers. Export selection handlers must not run until
+                // InitializeComponent has created the complete checkbox tree.
                 var page = new BackupExportSettingsPage();
                 if (!page.IsInitialized)
                     throw new InvalidOperationException("BackupExportSettingsPage did not finish WPF initialization.");
 
-                // All three export checkboxes default to IsChecked="True" in XAML, so
-                // UpdateExportButtonState should already have computed HasAnySection == true
-                // and left the export button enabled. This only holds if `_initialized`
-                // was set correctly by the time the constructor finished.
-                Assert(page.ExportButton.IsEnabled,
-                    "expected ExportButton to be enabled after construction (all sections checked)");
-
-                // Exercise the guard again post-construction: unchecking every export
-                // section should disable the button, and re-checking one should re-enable
-                // it. This fails if `_initialized` is ever re-inverted (e.g.
-                // `if (_initialized) return;`) or never set, since UpdateExportButtonState
-                // would then never run and the button would stay stuck at its initial state
-                // regardless of checkbox changes.
+                // Post-construction changes prove handlers are attached and state is
+                // recomputed; the button's default enabled value alone proves nothing.
                 page.ExportSettingsCheckbox.IsChecked = false;
                 page.ExportModesCheckbox.IsChecked = false;
                 page.ExportVocabularyCheckbox.IsChecked = false;
--- a/app/windows/HyperWhisper/Views/Pages/Settings/BackupExportSettingsPage.xaml.cs
+++ b/app/windows/HyperWhisper/Views/Pages/Settings/BackupExportSettingsPage.xaml.cs
@@ -13,16 +13,6 @@ public partial class BackupExportSettingsPage : Page
     /// <summary>Path of the backup file currently staged for selective import.</summary>
     private string? _pendingImportPath;
 
-    /// <summary>
-    /// True once the constructor has fully finished. A CheckBox's IsChecked="True" in
-    /// XAML fires Checked/ExportSection_Changed *during* InitializeComponent, before
-    /// sibling checkboxes further down the tree have been created — guard on this
-    /// explicit flag (set at the very end of the constructor) rather than on WPF's own
-    /// IsInitialized, whose exact timing relative to nested BAML parsing isn't a
-    /// documented contract we want to depend on.
-    /// </summary>
-    private bool _initialized;
-
     public BackupExportSettingsPage()
     {
         InitializeComponent();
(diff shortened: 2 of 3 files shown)


ray-amjad:

**Codex:** clean, no `[P1]`/blocking findings — affirms the `_initialized` guard fix is correct.

**Claude (medium tier):** 2 CONFIRMED findings, both non-blocking:

1. `app/windows/HyperWhisper.SmokeTests/Program.cs:355` — The smoke test meant to guard this exact regression still asserts WPF's own `page.IsInitialized` property, not the new `_initialized` field the fix actually depends on. If a future edit moves `_initialized = true;` above `InitializeComponent()`, forgets to set it, or re-inverts the guard, this test keeps passing while the original construction-order NRE silently returns. **Codex independently converged on this same gap** (noted it wasn't blocking but flagged the same test-coverage hole).

2. `app/windows/HyperWhisper/Views/Pages/Settings/BackupExportSettingsPage.xaml.cs:24` — This is the only file among the 13 in `Settings/` that doesn't use the established `Loaded += OnLoaded` deferred-initialization convention (all 12 siblings use it). The PR introduces a second, ad-hoc idiom (`_initialized` latch) for the same problem instead of following the existing pattern.

No blocking issues found by either reviewer — fixing #1 (test coverage) in this round; #2 is a convention note.

ray-amjad:

**Codex:** clean, zero P1/P2/blocking findings — confirms the guard and the strengthened smoke test are correct with no regression.

**Claude (medium tier):** 2 CONFIRMED, both moderate/non-blocking:

1. `app/windows/HyperWhisper.SmokeTests/Program.cs:369-370` — The *first* new assertion (`ExportButton.IsEnabled` right after construction) is vacuous: `Button.IsEnabled` defaults to `true` in WPF regardless of whether `_initialized` guard logic runs at all, so this specific assertion would pass even if the fix were fully reverted. The *later* assertions (uncheck-all → expect disabled, re-check-one → expect enabled again) do correctly catch a broken guard, so the test as a whole still has real coverage — just that one line is misleading.
2. Same convention note as round 1: `BackupExportSettingsPage.xaml.cs`'s `_initialized` latch is a bespoke one-off pattern; 11 of 12 sibling Settings pages avoid this whole bug class via `Loaded`-deferred setup instead. Not incorrect, just inconsistent with established convention.

This is the 2nd review round (the cap) — per policy, stopping here rather than looping further. Neither finding is blocking; noting both as known follow-ups rather than fixing in a 3rd round.

app/windows/HyperWhisper/Views/Pages/Settings/BackupExportSettingsPage.xaml.cs
```diff
@@ -36,11 +47,9 @@ private void ExportSection_Changed(object sender, RoutedEventArgs e)
 
     private void UpdateExportButtonState()
     {
-        // Setting a checkbox's IsChecked in XAML fires Checked/ExportSection_Changed
-        // *during* InitializeComponent, before the sibling checkboxes below it have
-        // been created. Bail out until the whole control tree is built so
+        // Bail out until the constructor has finished (see _initialized) so
         // BuildExportSelection never dereferences a not-yet-created checkbox.
-        if (!IsInitialized)
+        if (!_initialized)
```
ray-amjad: **This change is semantically a no-op — it almost certainly does not fix HYPERWHISPER-TK.**

`InitializeComponent()` calls `Application.LoadComponent(this, uri)`. The XAML object writer calls `BeginInit()` on the root `Page` before it parses any children and `EndInit()` only after the whole tree is built; `FrameworkElement.EndInit()` is what sets `IsInitialized = true`. So `this.IsInitialized` is **false** for the entire duration of BAML parsing — exactly the window in which `IsChecked="True"` fires `ExportSection_Changed`. The old `if (!IsInitialized) return;` already covered that window completely.

The only extra coverage `_initialized` buys is the region between the end of `InitializeComponent()` and the end of the constructor, and the only code there is:

```csharp
var vocabCount = VocabularyService.Instance.GetAll().Count;
ExportVocabularyCheckbox.Content = Loc.S(...);
```

Neither raises `Checked`/`Unchecked`, so `UpdateExportButtonState()` cannot be re-entered in that region. Old and new guard are equivalent for every reachable code path.

That matters because the guard is already in `main` (it is the line this PR deletes). If the Sentry issue is still reporting `BuildExportSelection` NREs from shipped builds that contain the `IsInitialized` guard, the root cause is somewhere else and this PR will not stop the crash. Worth confirming before merge: which app version(s) do the 31 events come from, and did they ship with the `IsInitialized` guard? If they did, the stack needs a second look (e.g. a null `ExportButton`/checkbox reference from a different entry point, or a `BuildExportSelection` call that is not routed through `UpdateExportButtonState`).

app/windows/HyperWhisper.SmokeTests/Program.cs
```diff
@@ -351,10 +351,40 @@ await ExpectAsync<OperationCanceledException>(() => RustRetry.PerformAsync(
...
                 var page = new BackupExportSettingsPage();
                 if (!page.IsInitialized)
                     throw new InvalidOperationException("BackupExportSettingsPage did not finish WPF initialization.");
 
+                // All three export checkboxes default to IsChecked="True" in XAML, so
+                // UpdateExportButtonState should already have computed HasAnySection == true
+                // and left the export button enabled. This only holds if `_initialized`
+                // was set correctly by the time the constructor finished.
+                Assert(page.ExportButton.IsEnabled,
+                    "expected ExportButton to be enabled after construction (all sections checked)");
```
ray-amjad: This assertion is vacuous, and its comment is wrong.

`Button.IsEnabled` defaults to `true`. If `_initialized` were never set (or `UpdateExportButtonState` never ran at all), `ExportButton.IsEnabled` would still be `true` here and this assert would still pass. So it does **not** verify that "`_initialized` was set correctly by the time the constructor finished" — it verifies nothing that the plain `new BackupExportSettingsPage()` line above does not already verify.

The two assertions below it (uncheck-all → disabled, re-check → enabled) are the ones that actually catch a never-set `_initialized`, since they require `UpdateExportButtonState` to have real effect. Either drop this one or make it non-vacuous by first forcing the opposite state, e.g.:

```suggestion
                page.ExportButton.IsEnabled = false;
                page.ExportSettingsCheckbox.IsChecked = false;
                page.ExportSettingsCheckbox.IsChecked = true;
                Assert(page.ExportButton.IsEnabled,
                    "expected ExportButton to be enabled after construction (all sections checked)");
```

Separately, the neighbouring comment claiming these post-construction asserts catch an inverted guard (`if (_initialized) return;`) is inaccurate: with the guard inverted the NRE fires inside `InitializeComponent`, so the constructor throws and execution never reaches these lines. The test still fails, just not for the stated reason.

ray-amjad:

---

### `app/windows/HyperWhisper/Views/Pages/Settings/BackupExportSettingsPage.xaml.cs:52` — ✅ **CONFIRMED**

**Original claim:** The `IsInitialized` → `_initialized` swap is semantically a no-op (IsInitialized is already false for the whole BAML parse), so the PR does not actually fix HYPERWHISPER-TK.

**What the code actually does:** Pre-PR file (git show e3708d4:.../BackupExportSettingsPage.xaml.cs, lines 37-47) already had the identical guard `if (!IsInitialized) return;` inside UpdateExportButtonState — the guard is NOT new, only its condition changed. The new guard opens at line 35 (`_initialized = true;`), the old one opened when the root Page's EndInit ran at the end of InitializeComponent (line 28). The only code between those two points is lines 31-33 (`VocabularyService.Instance.GetAll().Count` and `ExportVocabularyCheckbox.Content = ...`); assigning ContentControl.Content cannot raise Checked/Unchecked, so no ExportSection_Changed can fire in the newly-covered window and BuildExportSelection (lines 58-67, which dereferences IncludeApiKeysCheckbox declared at BackupExportSettingsPage.xaml:73, i.e. after the three IsChecked="True" checkboxes at xaml:46-56) is reached in exactly the same set of states as before. Empirical proof the old guard worked: the pre-existing smoke test at app/windows/HyperWhisper.SmokeTests/Program.cs:361-363 (added 2026-07-05 in e290a2a) does `new BackupExportSettingsPage()` and then asserts `page.IsInitialized` is true; Run() at Program.cs:416-428 turns any thrown NRE into a failure and a non-zero exit, and .github/workflows/windows-release.yml:187 runs that smoke test before building installers — yet Windows v1.8.1 (8d71b65, 2026-07-21) and v1.8.2 (88c0ff6, 2026-07-29) both shipped. So construction with only the IsInitialized guard demonstrably does not NRE, and IsInitialized demonstrably flips true by the end of InitializeComponent. The page is only ever created via `new` (app/windows/HyperWhisper/Views/Pages/SettingsPage.xaml.cs:66), never nested in another XAML, so there is no alternate parse path where IsInitialized could be true while a checkbox field is still null.

**Corrected statement:** The diff is behavior-preserving. Caveat on severity/wording: it is a defensible readability/robustness hardening (an explicit local invariant instead of relying on WPF-internal EndInit timing), not a wrong change — but it fixes no reachable crash. Since main's code already cannot produce the reported NRE, the Sentry issue must originate from a build predating the guard (or from a different path); merging this PR will not move the Sentry issue.


### `app/windows/HyperWhisper.SmokeTests/Program.cs:369` — 🟡 **PARTLY TRUE**

**Original claim:** Vacuous assert: Button.IsEnabled defaults to true, so `Assert(page.ExportButton.IsEnabled, ...)` passes even if `_initialized` is never set or UpdateExportButtonState never runs.

**What the code actually does:** The premise holds for that one line: UIElement.IsEnabled defaults to true and PrimaryButtonStyle (app/windows/HyperWhisper/Themes/Generic.xaml:99-122) has no IsEnabled setter — its only IsEnabled trigger sets the template Border's Opacity — and the page is never put in a window, so no ancestor coerces it false. If UpdateExportButtonState always early-returned, ExportButton.IsEnabled would still be true and the assert at Program.cs:369-370 would pass. But the claim's consequence ("the test does not catch a never-set _initialized") is wrong: the very next assertions in the same added block do catch it. Program.cs:378-382 unchecks all three export checkboxes and asserts `!page.ExportButton.IsEnabled`; CheckBox raises Unchecked on that assignment regardless of tree membership, so with a broken/never-set `_initialized` the button would remain enabled and Assert (Program.cs:434-438, throws InvalidOperationException) would fail the Run() block. Program.cs:384-386 then re-checks one box and asserts re-enable, catching an inverted guard.

**Corrected statement:** Only the first of the three new assertions is non-discriminating; the added test block as a whole is not vacuous — Program.cs:381 and Program.cs:385 do fail if `_initialized` is never set or the guard is inverted. This is a minor test-quality nit (the line-369 assert could be strengthened by pre-setting ExportButton.IsEnabled = false before construction, or dropped), not a coverage gap.


---

PR #50, opened 2026-07-17
Fix launch-time license validation network resilience (HYPERWHISPER-F4)

ray-amjad:
## Problem

Sentry HYPERWHISPER-F4: License validation network error — 419 occurrences since January, 26 users, component `license`, endpoint `/api/license/validate`. Sample event: app launch at 05:08:35, error at 05:08:38 (~3s post-launch) — consistent with launch-time license validation racing a network that isn't up yet (wake-from-sleep, captive portal, DNS not resolved yet).

## Investigation

- A cached "last known good" license verdict + 7-day offline grace already existed (`hw-license` cache, unmodified by PR #35), and `LicenseNetworkService` already retried and fell back to cache silently on exhaustion — so launch never hard-failed to the user.
- The gap: launch-time revalidation reused the same `.cloud` retry preset as explicit user-triggered "Activate" (3 attempts, worst case ~30s). Since this runs silently in the background at every launch, a flaky-network user could sit on stale `.trial` status for up to 30s before cache fallback kicked in — matching the Sentry timing.
- PR #35 fixed a different, unrelated bug (validation cache not keyed to the license key, causing a rejected replacement key to clobber a valid cached status). Confirmed via its diff — it doesn't touch network-failure resilience.

## Fix

- New `RetryConfiguration.licenseLaunchValidation` — a short, bounded retry preset (3 attempts, ~2s total backoff) used only for launch-time validation, distinct from the `.cloud` preset used by explicit user-triggered activation (unchanged).
- `LicenseNetworkService.validateLicense(_:isLaunchValidation:)` classifies failures via `isNetworkFailure(_:)` (connectivity-only URLErrors) vs. real server verdicts, and now emits a distinct, non-noisy Sentry signal (`component: license, reason: offline_no_cache`) only when retries are exhausted *and* there's no cached verdict to fall back on. Successful cache-fallback recovery stays silent by design — no user-facing error, no Sentry noise.
- `LicenseManager.loadStoredLicense()` now flags its call as launch validation so it gets the tighter budget.
- New unit tests for the retry preset bounds and the network-failure classifier.

## Hard constraint honored

Entitlement enforcement is untouched and stays fully server-side — no client-side bypass, no fake/test keys, no weakening of the actual license check. This only changes how the client tolerates *not yet being able to reach the server* at launch, by trusting a verdict the server already gave it earlier (the existing cache) instead of surfacing an error.

## Out of scope (flagging, not fixing here)

Same Sentry event's logs show an unrelated Core Data migration failure (`NSCocoaErrorDomain 134100`, store model version incompatible) — supports the ongoing V3/V4 Core Data migration investigation. Not touched in this PR.

Refs HYPERWHISPER-F4.

---

### DIFF
--- a/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
+++ b/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
@@ -109,6 +109,26 @@ struct NetworkConfig {
     /// point of the tighter launch-time budget. HYPERWHISPER-F4.
     static let licenseLaunchValidationTimeout: TimeInterval = 2.5
 
+    /// Delay before `LicenseManager` retries launch-time license validation
+    /// again in the background, after the first attempt exhausted its retries
+    /// due to a genuine connectivity failure and fell back to a cached (or
+    /// Invalid) verdict.
+    ///
+    /// The tight `licenseLaunchValidationTimeout` + `.licenseLaunchValidation`
+    /// retry budget above is a deliberate trade-off: it self-heals quickly from
+    /// a truly dead network, but it can also misclassify a merely-SLOW-but-live
+    /// network (weak wifi, VPN overhead) as a network failure, since a request
+    /// that's still hanging past ~2.5s is treated as stuck rather than "about to
+    /// succeed." Without a prompt follow-up, a legitimately-licensed user who
+    /// hit that misclassification would ride the cached fallback verdict for up
+    /// to the full 24h cache TTL / 7-day offline grace before the app tries the
+    /// real server again. This short one-shot retry closes that gap: by the
+    /// time it fires, a merely-slow connection has almost certainly stabilized,
+    /// so the app promptly re-confirms against the real server instead of
+    /// silently trusting a stale-but-still-"valid"-looking cache for up to a
+    /// week. HYPERWHISPER-F4 (review round 2).
+    static let licenseLaunchValidationRetrySoonDelay: TimeInterval = 45
+
     // MARK: - Caching and Retry Configuration
     
     /// Duration to cache successful license validation (24 hours)
--- a/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
+++ b/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
@@ -98,15 +98,11 @@ struct NetworkConfig {
     /// Per-request timeout for the silent, launch-time license revalidation
     /// (`LicenseNetworkService.validateLicense(_:isLaunchValidation: true)`).
     ///
-    /// Deliberately much shorter than `licenseValidationTimeout`: this call races
-    /// a network that may not be up yet (wake-from-sleep, captive portal, DNS not
-    /// resolved yet), so a request that's still hanging several seconds in is
-    /// almost certainly stuck on a dead network, not about to succeed — there's
-    /// no benefit to holding the full 10s per attempt. Paired with the short
-    /// `.licenseLaunchValidation` backoff, this keeps the worst case (all 3
-    /// attempts time out) to single-digit seconds instead of the ~30s that
-    /// `licenseValidationTimeout` × 3 attempts would otherwise cost — the whole
-    /// point of the tighter launch-time budget. HYPERWHISPER-F4.
+    /// Deliberately much shorter than `licenseValidationTimeout`. This is used
+    /// only when launch-time revalidation has a cached server verdict within the
+    /// offline grace period. If no usable cache exists, launch validation keeps
+    /// the normal timeout so a slow-but-live network is not mistaken for an
+    /// invalid license.
     static let licenseLaunchValidationTimeout: TimeInterval = 2.5
 
     /// Delay before `LicenseManager` retries launch-time license validation
--- a/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
+++ b/app/macos/hyperwhisper/Configuration/NetworkConfig.swift
@@ -94,7 +94,21 @@ struct NetworkConfig {
     /// Timeout for license validation calls (in seconds)
     /// Shorter timeout to prevent blocking UI
     static let licenseValidationTimeout: TimeInterval = 10
-    
+
+    /// Per-request timeout for the silent, launch-time license revalidation
+    /// (`LicenseNetworkService.validateLicense(_:isLaunchValidation: true)`).
+    ///
+    /// Deliberately much shorter than `licenseValidationTimeout`: this call races
+    /// a network that may not be up yet (wake-from-sleep, captive portal, DNS not
+    /// resolved yet), so a request that's still hanging several seconds in is
+    /// almost certainly stuck on a dead network, not about to succeed — there's
+    /// no benefit to holding the full 10s per attempt. Paired with the short
+    /// `.licenseLaunchValidation` backoff, this keeps the worst case (all 3
+    /// attempts time out) to single-digit seconds instead of the ~30s that
+    /// `licenseValidationTimeout` × 3 attempts would otherwise cost — the whole
+    /// point of the tighter launch-time budget. HYPERWHISPER-F4.
+    static let licenseLaunchValidationTimeout: TimeInterval = 2.5
+
     // MARK: - Caching and Retry Configuration
     
     /// Duration to cache successful license validation (24 hours)
--- a/app/macos/hyperwhisper/Managers/LicenseManager.swift
+++ b/app/macos/hyperwhisper/Managers/LicenseManager.swift
@@ -99,18 +99,27 @@ class LicenseManager: ObservableObject {
     }
 
     /// Validates a license key with the backend.
-    func validateLicense(_ licenseKey: String) async -> LicenseValidationResult {
+    /// - Parameter isLaunchValidation: forwarded to `LicenseNetworkService` — `true`
+    ///   only for the silent background revalidation `loadStoredLicense()` fires at
+    ///   launch, selecting its tighter retry budget. See HYPERWHISPER-F4.
+    func validateLicense(_ licenseKey: String, isLaunchValidation: Bool = false) async -> LicenseValidationResult {
         isValidating = true
         lastError = nil
         defer { isValidating = false }
 
-        let result = await networkService.validateLicense(licenseKey)
+        let result = await networkService.validateLicense(licenseKey, isLaunchValidation: isLaunchValidation)
         await processValidationResult(result)
 
         return result
     }
 
     /// Loads stored license from UserDefaults, revalidates if cache expired (24h).
+    ///
+    /// The revalidation call is tagged `isLaunchValidation: true` so a stale
+    /// network at launch (wake-from-sleep, captive portal, DNS not up yet) gets a
+    /// short bounded retry and — if that's still not enough — falls back to the
+    /// last cached server verdict instead of leaving `licenseStatus` stuck at its
+    /// default `.trial` for the whole `.cloud` retry budget. HYPERWHISPER-F4.
     func loadStoredLicense() async {
         guard let storedKey = networkService.getStoredLicenseKey() else {
             licenseStatus = .trial
@@ -118,7 +127,7 @@ class LicenseManager: ObservableObject {
         }
 
         if networkService.shouldRevalidateLicense() {
-            _ = await validateLicense(storedKey)
+            _ = await validateLicense(storedKey, isLaunchValidation: true)
         } else if let cachedStatus = networkService.getCachedLicenseStatus() {
             licenseStatus = cachedStatus
         }
--- a/app/macos/hyperwhisper/Managers/LicenseManager.swift
+++ b/app/macos/hyperwhisper/Managers/LicenseManager.swift
@@ -59,6 +59,14 @@ class LicenseManager: ObservableObject {
     /// Network service for license API calls
     private let networkService: LicenseNetworkService
 
+    /// One-shot background retry, scheduled only when launch-time validation
+    /// fell back to a cached verdict because of a genuine network failure (see
+    /// `loadStoredLicense()`). Held so a second `loadStoredLicense()` call
+    /// (unlikely in practice — `LicenseManager` is a long-lived singleton — but
+    /// possible in tests) cancels any still-pending retry rather than stacking
+    /// them. HYPERWHISPER-F4 (review round 2).
+    private var networkFailureRetryTask: Task<Void, Never>?
+
     // MARK: - Initialization
 
     init() {
@@ -120,6 +128,11 @@ class LicenseManager: ObservableObject {
     /// short bounded retry and — if that's still not enough — falls back to the
     /// last cached server verdict instead of leaving `licenseStatus` stuck at its
     /// default `.trial` for the whole `.cloud` retry budget. HYPERWHISPER-F4.
+    ///
+    /// If that fallback happened specifically because the network couldn't be
+    /// reached (`result.networkFailureFallback`), a short background retry is
+    /// scheduled — see `scheduleRetrySoonAfterNetworkFallback`. HYPERWHISPER-F4
+    /// (review round 2).
     func loadStoredLicense() async {
         guard let storedKey = networkService.getStoredLicenseKey() else {
             licenseStatus = .trial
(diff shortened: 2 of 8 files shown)


ray-amjad:
## Review round 1 — Claude (medium tier) + Codex

**Entitlement check:** both reviewers confirmed no client-side bypass, fake/test key, or weakening of the license check — the Rust core still owns all validate/cache/grace decisions unchanged.

**Findings (2 flagged by both reviewers, 2 Claude-only):**

1. **[Both — P1] Launch retry preset doesn't actually bound the delay it targets** — `LicenseNetworkService.swift`. The new preset only shortens the backoff *sleep* between attempts; it doesn't shorten the per-request `licenseValidationTimeout` (10s). So 3 attempts can still each hang for up to 10s, leaving `licenseStatus` at `.trial` for ~30s in the worst case — not the couple-second self-heal window the PR describes. Needs a launch-specific, much shorter per-request timeout, not just shorter backoff.

2. **[Both — P2] `offline_no_cache` Sentry signal mislabels real cached verdicts** — `LicenseNetworkService.swift` (~L239). The branch decides "no usable cache" by checking `outcome.isValid`, but a genuinely cached `Expired`/`Invalid` verdict (still in the 7-day grace window) is a real cache hit, not an absence of one. Every offline user with a cached-but-non-active status will fire a false "no cache" alarm, diluting the new signal's whole purpose.

3. **[Claude-only] Duplicated network-failure classifier** — new `isNetworkFailure(_:)` re-implements connectivity-error detection that already exists (and has already drifted) in `TranscriptionPipeline+ErrorClassification.swift` and `HyperWhisperCloudProvider.isDnsError`. Should reuse one of the existing classifiers rather than adding a third.

4. **[Claude-only, minor] Doc comment inaccuracy** — `Retrying.swift` claims "~2s of total backoff sleep" for the new preset; actual worst case is 1.5–1.8s since there's no sleep after the final attempt.

Fixing 1–4 now.

ray-amjad:
## Review round 2 — Claude (high tier) + Codex

**Entitlement check:** confirmed clean again — no bypass/fake-key risk, existing `validateLicense` call sites unaffected by the new default parameter.

**Codex's P1 (actor isolation)** — investigated in depth and **refuted**: `transientURLErrorCodes` is a `static let` of a `Sendable` value type; Swift's actor-isolation rules exempt immutable static constants from `@MainActor` inference (unlike static *functions*, which do inherit isolation — this file's own `sentryFingerprintForTranscriptionFailure` needed an explicit `nonisolated`, the `static let` didn't). Backed by `SWIFT_VERSION = 5.0` with no strict-concurrency opt-in anywhere in the project, and the same `static let`-in-`@MainActor`-class pattern already exists elsewhere in the codebase (`SettingsManager`, `BackupManager`). No compile break expected.

**Findings from Claude (high tier), most severe first:**

1. **[CONFIRMED]** `NetworkConfig.swift:110` — the new 2.5s timeout + 3-attempt retry (~9s worst case) can misclassify a real-but-slow network as dead, sending a legitimately-licensed user into the cached-verdict fallback path, which can then persist for up to 7 days.
2. **[CONFIRMED]** `LicenseNetworkService.swift:291` — the new `offline_no_cache` Sentry signal computes `isNetworkFailure(error)` but never actually uses it to distinguish real connectivity failure from a persistent server 5xx/429 — both get tagged identically, muddying the new signal's diagnostic value.
3. **[CONFIRMED]** `TranscriptionPipeline+ErrorClassification.swift:132` — doc comment calls `transientURLErrorCodes` "the canonical set," but `classifyTranscriptionError` still maintains its own separate, narrower 6-code list nearby, unreused — comment overstates what's actually shared.
4. **[PLAUSIBLE]** `LicenseNetworkService.swift:261` — redundant, non-atomic double FFI read (`licenseOfflineFallbackOutcome` + `licenseCachedStatusWithinGrace`), narrow theoretical race against concurrent `clearStoredLicense()`.
5. **[PLAUSIBLE]** `LicenseNetworkService.swift:143` — timeout and retry-config selection done via two independent ternaries on the same bool ~18 lines apart, must be kept in lockstep by hand.
6. **[PLAUSIBLE, pre-existing, out of scope]** `BackupManager.swift:68` — post-restore license revalidation uses the old, wider timeout budget too; architecturally identical to the launch-time gap but this is a pre-existing pattern this PR didn't introduce — not fixing here.

Fixing 1–5 now (skipping 6, out of scope/pre-existing). This will be the last review round per the loop cap — any remainder after this fix will be reported honestly rather than looped again.

app/macos/hyperwhisper/Models/LicenseValidationResult.swift
```diff
@@ -59,4 +59,23 @@ struct LicenseValidationResult {
...
+    /// `getCachedLicenseStatus()` branch), where no revalidation is needed.
+    /// `LicenseManager` uses this flag to schedule a short, one-shot
+    /// background retry after a launch-time validation falls back this way —
+    /// a merely-slow-but-live network (weak wifi, VPN overhead) shouldn't
+    /// leave a legitimately-licensed user riding a stale cached verdict for up
+    /// to the full 7-day offline grace period. HYPERWHISPER-F4 (review round 2).
+    ///
+    /// Defaults to `false`; only `LicenseNetworkService.validateLicense`'s
+    /// offline-fallback catch branch sets it `true`.
+    let networkFailureFallback: Bool = false
```
ray-amjad: **Build break — this does not compile, and the flag can never be `true`.**

In Swift, a `let` property **with a default value is excluded from the memberwise initializer** (it is already initialized and cannot be assigned). So `LicenseNetworkService.adapt(_:networkFailureFallback:)`, which does `LicenseValidationResult(..., networkFailureFallback: networkFailureFallback)`, fails to compile. CI confirms it:

```
LicenseNetworkService.swift:489:37: error: extra argument 'networkFailureFallback' in call
```
(`Build & Test (macOS, unsigned)` is red on this head.)

Even setting the compile error aside, the semantics are wrong: `networkFailureFallback` is hardcoded `false` for every instance, so
- `LicenseManager.loadStoredLicense()`'s `if result.networkFailureFallback` is dead code and the entire review-round-2 retry-soon feature never fires, and
- the new test `adaptPropagatesNetworkFailureFallbackFlagWhenSet` (`#expect(fellBackDueToNetwork.networkFailureFallback)`) asserts something that can never hold.

`var` with a default value *is* included in the memberwise init (as a defaulted parameter), which keeps every existing call site that omits the argument working:

```suggestion
    var networkFailureFallback: Bool = false
```

app/macos/hyperwhisper/Managers/LicenseManager.swift
```diff
@@ -99,31 +107,81 @@ class LicenseManager: ObservableObject {
...
+    /// case. HYPERWHISPER-F4 (review round 2).
+    private func scheduleRetrySoonAfterNetworkFallback(licenseKey: String) {
+        networkFailureRetryTask?.cancel()
+        networkFailureRetryTask = Task { [weak self] in
+            let delayNanoseconds = UInt64(NetworkConfig.licenseLaunchValidationRetrySoonDelay * 1_000_000_000)
+            try? await Task.sleep(nanoseconds: delayNanoseconds)
+            guard let self, !Task.isCancelled else { return }
+
+            AppLogger.network.info("License validation · retrying soon after launch-time network-failure fallback")
+            _ = await self.validateLicense(licenseKey, isLaunchValidation: true)
```
ray-amjad: **The retry-soon task captures a stale license key and can clobber a freshly-activated license.**

`licenseKey` is captured by value at schedule time and replayed 45s later, with no check that it is still the stored key.

Scenario:
1. Launch on a dead/slow network → `loadStoredLicense()` falls back to cache, schedules retry-soon with `OLD-KEY`.
2. Within the next 45s the user opens Settings and activates `NEW-KEY`. `activateLicense` succeeds (200), `licenseStatus == .active`, stored key is now `NEW-KEY`.
3. At t=45s this task fires and validates `OLD-KEY`. The server answers 200 with an invalid/expired verdict for that key → `validateLicense` → `processValidationResult` sets `licenseStatus = .invalid` and `lastError`.

A user who just activated a valid key is now shown as invalid. (The Rust cache itself is protected — `licensePersistValidationVerdict` only updates the cache when `attemptedKey == storedKey` — but the in-memory `@Published` state is not.)

Re-reading the stored key at fire time fixes it:

```suggestion
            guard networkService.getStoredLicenseKey() == licenseKey else { return }
            AppLogger.network.info("License validation · retrying soon after launch-time network-failure fallback")
            _ = await self.validateLicense(licenseKey, isLaunchValidation: true)
```

(`getStoredLicenseKey()` is already used by `loadStoredLicense()`; it may need to be reachable from here.)

app/macos/hyperwhisper/Managers/LicenseManager.swift
```diff
@@ -99,31 +107,81 @@ class LicenseManager: ObservableObject {
...
+        }
+    }
+
     /// Clears stored license and resets to the unlicensed (trial) state.
     func clearLicense() {
```
ray-amjad, line 228: **`clearLicense()` / `deactivateLicense()` do not cancel `networkFailureRetryTask`, so deactivation can be undone 45s later.**

Scenario:
1. Launch offline → retry-soon scheduled with the stored key.
2. Within 45s the user hits Deactivate. `deactivateLicense()` → `clearLicense()` → stored key wiped, `licenseStatus = .trial`, `lastError = nil`.
3. At t=45s the pending task calls `validateLicense(oldKey, isLaunchValidation: true)`. Still offline, so `LicenseNetworkService`'s catch branch hits the `licenseStoredLicenseKey(store:) != trimmedKey` guard (stored key is now nil) and returns `status: .invalid, errorMessage: "Unable to verify license while offline"`.
4. `processValidationResult` sets `licenseStatus = .invalid` and `lastError` to that message.

The user who just deactivated now sees an *invalid license* error instead of the trial state they asked for. Cancelling the pending task on clear fixes it:

```suggestion
    func clearLicense() {
        networkFailureRetryTask?.cancel()
        networkFailureRetryTask = nil
        networkService.clearStoredLicense()
```

app/macos/hyperwhisper/Configuration/NetworkConfig.swift
```diff
@@ -94,7 +94,41 @@ struct NetworkConfig {
...
+    /// Deliberately much shorter than `licenseValidationTimeout`: this call races
+    /// a network that may not be up yet (wake-from-sleep, captive portal, DNS not
+    /// resolved yet), so a request that's still hanging several seconds in is
+    /// almost certainly stuck on a dead network, not about to succeed — there's
+    /// no benefit to holding the full 10s per attempt. Paired with the short
+    /// `.licenseLaunchValidation` backoff, this keeps the worst case (all 3
+    /// attempts time out) to single-digit seconds instead of the ~30s that
+    /// `licenseValidationTimeout` × 3 attempts would otherwise cost — the whole
+    /// point of the tighter launch-time budget. HYPERWHISPER-F4.
+    static let licenseLaunchValidationTimeout: TimeInterval = 2.5
```
ray-amjad, line 106: **2.5s makes the no-cache path *more* reachable, and that path shows a paying user an invalid-license error.**

The PR frames the trade-off as "worst case a slow network rides the cache instead of the live verdict", which is safe when a cache exists. But when there is **no** usable cache, `licenseOfflineFallbackOutcome` returns `Invalid`, and `LicenseManager.processValidationResult` then sets `licenseStatus = .invalid` and `lastError` — a visible "unable to verify license" for a legitimately licensed user.

Concrete case: user restores from a backup (`BackupManager` imports `DefaultsKey.licenseKey`, but there is no cached verdict), first launch is on a corporate VPN / weak hotspot where the round-trip takes ~4s. Before this PR each of the 3 attempts had 10s and would have succeeded; now all 3 abort at 2.5s and the user is marked invalid. Same for anyone whose cache has aged past the 7-day grace.

The intended mitigation for exactly this (the 45s retry-soon) is currently dead code — see the `networkFailureFallback` memberwise-init issue in `LicenseValidationResult.swift`. Once that is fixed the window shrinks to 45s, but it may still be worth either raising this to ~5s or skipping the tight budget when there is no cached verdict to fall back on.

app/macos/hyperwhisperTests/LicenseNetworkResilienceTests.swift
```diff
@@ -0,0 +1,286 @@
...
+        // self-heal without leaving a paying user looking unlicensed for long.
+        let totalBackoff = (1..<config.maxAttempts).reduce(0.0) { sum, attempt in
+            sum + config.delay(for: attempt)
+        }
+        #expect(totalBackoff < 5.0)
+
+        // Every single delay (even accounting for jitter) stays well under the
+        // `.cloud` budget's own per-attempt ceiling.
+        for attempt in 1..<config.maxAttempts {
+            #expect(config.delay(for: attempt) <= config.maxDelay)
```
ray-amjad: Vacuous assertion — `RetryConfiguration.delay(for:)` ends in `min(baseDelay + jitter, maxDelay)`, so `delay(for:) <= maxDelay` holds by construction for *every* configuration and can never fail. It would pass identically with `maxDelay: 300`.

If the intent is "each individual delay stays short", assert against a literal instead, e.g. `#expect(config.delay(for: attempt) <= 2.0)`.

ray-amjad:

**Result: 2 confirmed, 3 partly true.**

---

### `app/macos/hyperwhisper/Models/LicenseValidationResult.swift:80` — ✅ **CONFIRMED**

**Original claim:** `let networkFailureFallback: Bool = false` is excluded from the synthesized memberwise initializer, so `LicenseNetworkService.adapt(_:networkFailureFallback:)` does not compile and the flag can never be true.

**What the code actually does:** LicenseValidationResult.swift:80 declares `let networkFailureFallback: Bool = false` — a constant stored property WITH a default, which Swift omits from the synthesized memberwise init. LicenseNetworkService.swift:481-490 declares `static func adapt(_ outcome: ValidationOutcome, networkFailureFallback: Bool = false)` and passes `networkFailureFallback: networkFailureFallback` into `LicenseValidationResult(...)`. There is no explicit init and no `extension LicenseValidationResult` anywhere in app/ (grep returned zero). The PR's own CI job (Build & Test (macOS, unsigned), run 29584537369 job 87898110362) fails with exactly one compile error: `app/macos/hyperwhisper/Managers/LicenseNetworkService.swift:489:37: error: extra argument 'networkFailureFallback' in call`. The test at LicenseNetworkResilienceTests.swift:249 calls the same API and would fail identically. Even if the call site were removed, `let ... = false` makes the flag permanently false, so `LicenseManager.swift:144`'s `if result.networkFailureFallback` is dead — the PR's headline round-2 feature never runs.


### `app/macos/hyperwhisperTests/LicenseNetworkResilienceTests.swift:80` — ✅ **CONFIRMED**

**Original claim:** Vacuous assertion: `delay(for:) <= maxDelay` holds by construction.

**What the code actually does:** Retrying.swift:87-91 — `func delay(for attempt: Int)` computes `baseDelay`, adds jitter, and returns `min(baseDelay + jitter, maxDelay)` (line 90). `#expect(config.delay(for: attempt) <= config.maxDelay)` at LicenseNetworkResilienceTests.swift:80 is therefore true for every possible input; it can never fail. Worse, the comment above it (lines 77-78) says it checks "well under the `.cloud` budget's own per-attempt ceiling", but it compares against `config.maxDelay` (the launch preset's own 2.0s, Retrying.swift:74), not `RetryConfiguration.cloud.maxDelay` (30.0, Retrying.swift:48). The intended cross-preset check does exist separately at test line 93, so only line 80 is dead weight.

**Corrected statement:** The assertion is vacuous and should either be deleted or rewritten as the comparison its own comment describes: `config.delay(for: attempt) <= RetryConfiguration.cloud.maxDelay`, or an absolute bound such as `<= 2.4` (initialDelay 0.5 × 2^1 × 1.2 jitter).


### `app/macos/hyperwhisper/Managers/LicenseManager.swift:181` — 🟡 **PARTLY TRUE**

**Original claim:** The 45s retry task captures the license key by value at schedule time and can clobber a license the user activates in the meantime.

**What the code actually does:** The mechanism is real. LicenseManager.swift:145 passes `storedKey` into `scheduleRetrySoonAfterNetworkFallback(licenseKey:)`; the escaping closure at lines 175-182 captures that String by value and, 45s later, calls `self.validateLicense(licenseKey, isLaunchValidation: true)` (line 181) with the stale key. Nothing re-reads `networkService.getStoredLicenseKey()` inside the task. If the stale key K1 still validates Active, hw-license `cache.rs:141-143` (`if is_valid { store_license_key(store, key); }`) rewrites the stored key back to K1, overwriting a newly activated K2. If K1 comes back Expired/Invalid, the stored key survives but `processValidationResult` (LicenseManager.swift:198-203) still unconditionally overwrites the @Published `licenseStatus`/`customerEmail` with K1's verdict, so the UI state for K2 is clobbered either way. HOWEVER: the retry is scheduled only under `if result.networkFailureFallback` (line 144), and that flag is hard-wired to `false` by LicenseValidationResult.swift:80 — and the file does not compile at all (finding 1). So on this branch as it stands the path is unreachable dead code.

**Corrected statement:** Accurate statement: this is a latent bug, not a live one. `scheduleRetrySoonAfterNetworkFallback` can never fire on this branch (the flag is `let ... = false`, and the target file fails to compile), so no user can hit the clobber today. The moment finding 1 is fixed and the retry becomes reachable, the stale-key capture at LicenseManager.swift:175-181 does become a real key-and-status clobber via cache.rs:141-143. Fix both together: re-read the stored key inside the task (and bail if it changed) rather than capturing it at schedule time.


### `app/macos/hyperwhisper/Managers/LicenseManager.swift:186` — 🟡 **PARTLY TRUE**

**Original claim:** clearLicense() / deactivateLicense() do not cancel networkFailureRetryTask, so a deactivation is undone 45s later.

**What the code actually does:** The omission is real: `clearLicense()` (LicenseManager.swift:186-193) clears the store and resets `licenseStatus`/`customerEmail`/`customerName`/`lastError`, but never touches `networkFailureRetryTask` (declared line 68); the only `.cancel()` in the file is at line 174 inside the scheduler itself. `deactivateLicense()` (lines 95-107) routes through `clearLicense()`, so it inherits the gap. The pending task's `guard let self, !Task.isCancelled` (line 178) therefore passes, and line 181 revalidates the deactivated key; a still-Active verdict re-persists it via hw-license `cache.rs:141-143` and `processValidationResult` (lines 198-203) flips `licenseStatus` back to `.active` — deactivation undone. BUT the same reachability gate applies: the task is only ever created under `if result.networkFailureFallback` (line 144), which is permanently false (LicenseValidationResult.swift:80) on a branch that does not compile.

**Corrected statement:** Accurate statement: a real missing-cancellation defect, but currently unreachable dead code — the retry task is never scheduled on this branch. It becomes a live "deactivation silently reverts 45s later" bug as soon as finding 1 is fixed. Add `networkFailureRetryTask?.cancel(); networkFailureRetryTask = nil` to `clearLicense()`.


### `app/macos/hyperwhisper/Configuration/NetworkConfig.swift:110` — 🟡 **PARTLY TRUE**

**Original claim:** Dropping the launch timeout to 2.5s makes the no-cache path MORE reachable, and that path shows a paying user an invalid-license error.

**What the code actually does:** Half of this is confirmed. NetworkConfig.swift:110 sets `licenseLaunchValidationTimeout = 2.5` vs `licenseValidationTimeout = 10` (line 96), and LicenseNetworkService.swift:410-420 (`requestPolicy`) applies it to every launch-time validation, so a slow-but-live network that would have completed in 3-9s now throws `URLError.timedOut` on all 3 attempts and lands in the catch branch (line 227). The error path is also real: with no cached verdict, hw-license `cache.rs:203-211` returns `status: Invalid, is_valid: false, error_message: "Offline and no cached license"`, LicenseManager.swift:199-202 sets `licenseStatus = .invalid` and `lastError`, and CloudAccountSettingsSection.swift:187 renders that string in a `licenseErrorBanner`. What's overstated is the linkage: reaching the error requires `hasCachedVerdict == false` (LicenseNetworkService.swift:279-292), i.e. the user has no cached server verdict inside the 7-day grace (cache.rs:194). Any paying user who validated in the last 7 days gets the `hasCachedVerdict` branch at line 292-301 — cached verdict served, no error, no Sentry, which is exactly the PR's intent. So the shorter timeout mostly makes the SILENT cached-fallback path more reachable, not the error path.

**Corrected statement:** Accurate statement: the 2.5s timeout does make the offline-fallback catch branch more reachable on slow-but-live networks, but for the overwhelming majority (any user with a cached verdict inside the 7-day grace) that produces a silent cached-verdict fallback with no error. The invalid-license banner only reaches users who additionally have no cached verdict within 7 days — and that Invalid-plus-error outcome is pre-existing behavior (cache.rs:203-211 is unchanged by this PR), not introduced here; the PR only shortens the time to reach it. Severity is materially lower than stated, though the 2.5s value is still worth revisiting since it converts 'slow network' into 'network failure' by design.


---

PR #51, opened 2026-07-17
fix(macos): revalidate license and retry once on transcribe 401/403

ray-amjad:
Fixes the root cause behind **HYPERWHISPER-T2** (`TranscriptionPipeline.transcribeWithDetails failed` — 42 users, 109 events, `error_class: auth`, `errorCode: 4`, stage `transcribe`, not retryable, ~505ms against HyperWhisper Cloud, seen on current 2.41.0).

## Verification

Sentry event details were provided directly in the dispatch thread (no Sentry API access in this environment). Verified the failure path by tracing the code — see below.

## Root cause

`HyperWhisperCloudProvider.transcribe()` reads the license identifier via `LicenseManager.getTranscriptionIdentifier()`, which is a synchronous read of an in-memory `@Published licenseStatus` cache. That cache is only revalidated against the server at app launch or once per 24h (`LicenseNetworkService.shouldRevalidateLicense`), plus a 7-day offline grace period on network failure.

HyperWhisper macOS is a long-running menu-bar app. If a license lapses, is rotated, or is revoked server-side mid-session, the client keeps sending the stale identifier as if it were still valid for up to 24h+. When the backend correctly rejects it with 401/403, the shared Rust core classifies this as `Unauthorized` and intentionally treats it as **terminal on the first attempt** (`is_retryable` excludes 401/403, by design — to avoid retry-storming a genuinely bad credential). Nothing upstream of that terminal classification ever distinguished "cache is just stale" from "credential is actually bad," so every one of these hard-fails straight to the user with no recovery path, matching the single ~505ms non-retried round trip in the Sentry data.

`errorCode: 4` is not a designed error code — `TranscriptionError` doesn't conform to `CustomNSError`, so this is just Swift's default enum→NSError ordinal bridging for `.unauthorized`, which is the only case that can produce `error_class: auth` for the HyperWhisper Cloud provider (no BYOK key concept applies here).

## Fix

On catching `TranscriptionError.unauthorized` from the `/transcribe` call, force one live `licenseManager.validateLicense(...)` revalidation and retry the transcribe request exactly once with the freshly confirmed identifier. This preserves the core's intentional no-blind-retry policy (a genuinely revoked/invalid license still fails fast on the retry) while fixing the much more common "the local cache just hadn't caught up yet" case.

Refactored the request-building/send logic into a small local `sendTranscribeRequest(identifier:isLicensed:)` closure inside `transcribe()` so the retry path reuses it instead of duplicating the request-building/parseError/onTransportError wiring.

## Not fixed here (flagged as follow-ups)

- The other call sites that also use `getTranscriptionIdentifier()` (`/post-process`, credit balance fetch, streaming) could hit the same stale-cache 401 — this PR only covers the reported `transcribeWithDetails`/`transcribe` stage.
- `TranscriptionError` doesn't conform to `CustomNSError`, so `errorCode` in Sentry extras is an unstable compiler-assigned ordinal rather than a stable, self-describing code. Worth fixing separately for future Sentry triage.

References: HYPERWHISPER-T2

### DIFF
--- a/app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
@@ -380,7 +380,7 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
         // =====================================================================
         // STEP 3: Perform request via the shared executor + core retry loop.
         // =====================================================================
-        let response = try await Self.performTranscribeRequestWithLicenseRecovery(
+        let requestResult = try await Self.performTranscribeRequestWithLicenseRecovery(
             identifier: identifier,
             isLicensed: isLicensed,
             send: sendTranscribeRequest,
@@ -389,8 +389,14 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
             },
             currentIdentifier: {
                 await licenseManager.getTranscriptionIdentifier()
+            },
+            refreshServerAuthCache: { refreshedIdentifier in
+                try await creditManager.refreshServerLicenseCache(for: refreshedIdentifier)
             }
         )
+        let response = requestResult.response
+        let successfulIdentifier = requestResult.identifier
+        let successfulIsLicensed = requestResult.isLicensed
         try throwIfCancelled()
 
         // Parse the success response via the core.
@@ -453,8 +459,8 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
                             session: session,
                             text: segment,
                             prompt: prompt,
-                            identifier: identifier,
-                            isLicensed: isLicensed,
+                            identifier: successfulIdentifier,
+                            isLicensed: successfulIsLicensed,
                             mode: mode
                         )
                         processed.append(output.trimmingCharacters(in: .whitespacesAndNewlines))
@@ -465,8 +471,8 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
                         session: session,
                         text: transcript.text,
                         prompt: prompt,
-                        identifier: identifier,
-                        isLicensed: isLicensed,
+                        identifier: successfulIdentifier,
+                        isLicensed: successfulIsLicensed,
                         mode: mode
                     )
                 }
@@ -521,15 +527,27 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
     /// the freshly resolved identifier still being licensed. A revoked/expired
     /// license must keep the original unauthorized failure instead of silently
     /// re-sending the audio against the guest/device wallet.
+    struct TranscribeRequestResult {
+        let response: HttpResponse
+        let identifier: String
+        let isLicensed: Bool
+    }
+
     static func performTranscribeRequestWithLicenseRecovery(
         identifier: String,
         isLicensed: Bool,
         send: (String, Bool) async throws -> HttpResponse,
         revalidate: (String) async -> LicenseValidationResult,
-        currentIdentifier: () async -> (identifier: String, isLicensed: Bool)
-    ) async throws -> HttpResponse {
+        currentIdentifier: () async -> (identifier: String, isLicensed: Bool),
+        refreshServerAuthCache: (String) async throws -> Void
+    ) async throws -> TranscribeRequestResult {
         do {
-            return try await send(identifier, isLicensed)
+            let response = try await send(identifier, isLicensed)
+            return TranscribeRequestResult(
+                response: response,
+                identifier: identifier,
+                isLicensed: isLicensed
+            )
         } catch let error as TranscriptionError {
             guard case .unauthorized = error, isLicensed else {
                 throw error
@@ -547,7 +565,21 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
                 throw error
             }
 
-            return try await send(refreshed.identifier, refreshed.isLicensed)
+            do {
+                try await refreshServerAuthCache(refreshed.identifier)
+            } catch is CancellationError {
+                throw CancellationError()
+            } catch {
+                AppLogger.network.warning("HyperWhisper Cloud server license-cache refresh failed · preserving unauthorized response")
+                throw error
+            }
+
+            let response = try await send(refreshed.identifier, refreshed.isLicensed)
+            return TranscribeRequestResult(
+                response: response,
+                identifier: refreshed.identifier,
+                isLicensed: refreshed.isLicensed
+            )
         }
     }
 
--- a/app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
@@ -296,87 +296,95 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
             ? RustCoreMapping.boostVocabularyTerms(from: vocabulary)
             : []
         let trimmedModelId = selectedModelId.isEmpty ? nil : selectedModelId
-
-        let params = RustCoreMapping.transcribeParams(
-            audioPath: audioURL.path,
-            audioMime: contentType,
-            language: language,
-            vocabulary: vocabTermsForCore,
-            baseURL: NetworkConfig.hyperwhisperCloudURL,
-            licenseKey: isLicensed ? identifier : nil,
-            deviceID: isLicensed ? nil : identifier,
-            routedProvider: accuracyTier.sttProvider,
-            routedModel: trimmedModelId,
-            routedDomain: transcriptionDomain
-        )
-
-        let baseRequest: HttpRequest
-        do {
-            baseRequest = try hyperwhisperCloudBuildTranscribeRequest(params: params)
-        } catch let err as HwTranscriptionError {
-            throw RustCoreMapping.mapTranscriptionError(err, providerName: "HyperWhisper Cloud")
-        }
-
-        // The core does not add the platform `mode` query param or `User-Agent`
-        // header (both are HW-Cloud-specific, not part of the shared contract).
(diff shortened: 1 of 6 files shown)


app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
```diff
@@ -386,7 +394,39 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
...
+            // 24h (`LicenseNetworkService.shouldRevalidateLicense`), plus a 7-day
+            // offline grace period. A license that lapses/rotates server-side
+            // mid-session keeps being sent as if still valid until that cache
+            // expires, and the core intentionally never retries a 401/403 blind
+            // (`hw-net` `is_retryable`) to avoid retry-storming a genuinely bad
+            // credential. So on `.unauthorized`, force one live revalidation and
+            // retry exactly once with the confirmed-fresh identifier — this fixes
+            // the "cache is just stale" case while still failing fast for an
+            // actually-revoked license.
+            guard case .unauthorized = mapped, isLicensed else {
```
ray-amjad: **The revalidate-and-retry block is unreachable — this PR does not fix HYPERWHISPER-T2.**

A 401/403 never arrives at this `catch`. The `catch let err as HwTranscriptionError` on line 390 only wraps `hyperwhisperCloudParseTranscribeResponse(resp: response)`, and `response` is whatever `RustRetry.perform` returned — which by contract is **only a 2xx** (`RustRetry.swift:109-111`, `if (200...299).contains(...) { return response }`).

For a non-2xx, `RustRetry.perform` gives up at `RustRetry.swift:132-136` and throws `enrichRateLimited(parseError(response), ...)` — i.e. a `TranscriptionError`, already mapped. So on a 401 the throw happens at line 383 (`var response = try await sendTranscribeRequest(...)`) and propagates straight out of `transcribe()`. It is (a) not an `HwTranscriptionError`, so the `catch` on 390 does not match it, and (b) thrown before line 388 is ever reached.

Even reaching line 390 with `.unauthorized` is impossible from the other direction: the core parser can only produce `Unauthorized` for a non-200 status. `shared-core-rs/crates/hw-net/src/providers/hyperwhisper_cloud.rs:220-231` returns `classify_status_json(...)` only when `resp.status != 200`; for a 200 the only error variants it can yield are `NoSpeech` and `Parse`. So `guard case .unauthorized = mapped` is dead in every scenario, and the guard silently rethrows the `NoSpeech`/`Parse` error exactly as before the PR.

**Repro:** license lapses server-side → `/transcribe` returns 401 → `nextRetry` says `.giveUp` → `RustRetry` throws `TranscriptionError.unauthorized(provider:)` from line 383 → user sees the same hard auth failure the Sentry issue reports. No revalidation, no retry.

**Fix:** the recovery has to wrap the *send*, not the parse. Roughly:

```swift
var response: HttpResponse
do {
    response = try await sendTranscribeRequest(identifier: identifier, isLicensed: isLicensed)
} catch let error as TranscriptionError {
    guard case .unauthorized = error, isLicensed else { throw error }
    AppLogger.network.warning("HyperWhisper Cloud transcribe unauthorized · forcing license revalidation and retrying once")
    _ = await licenseManager.validateLicense(identifier)
    let (retryIdentifier, retryIsLicensed) = await licenseManager.getTranscriptionIdentifier()
    response = try await sendTranscribeRequest(identifier: retryIdentifier, isLicensed: retryIsLicensed)
}
try throwIfCancelled()
```

with the parse below left as a plain `let transcript = try ...` as on `main`.

app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift
```diff
@@ -386,7 +394,39 @@ class HyperWhisperCloudProvider: TranscriptionProvider {
...
+                throw mapped
+            }
+
+            AppLogger.network.warning("HyperWhisper Cloud transcribe unauthorized · forcing license revalidation and retrying once")
+            _ = await licenseManager.validateLicense(identifier)
+            let (retryIdentifier, retryIsLicensed) = await licenseManager.getTranscriptionIdentifier()
+
+            let retryResponse = try await sendTranscribeRequest(identifier: retryIdentifier, isLicensed: retryIsLicensed)
```
ray-amjad: (Applies once the placement issue above is fixed and this path can actually run.)

**The retry silently downgrades a licensed user to the guest/device-ID wallet instead of surfacing the auth failure.**

`validateLicense` result is discarded (`_ =` on line 415), and `getTranscriptionIdentifier()` (`LicenseManager.swift:191-198`) returns `(deviceID, false)` for anything other than `licenseStatus == .active`. So when revalidation confirms the license is genuinely revoked/expired — the exact case the comment says should "fail fast" — `retryIsLicensed` comes back `false` and line 418 re-sends the audio as an **unlicensed device-ID request**.

Scenario: user's subscription lapses mid-session → 401 → forced revalidation returns `.expired` → retry goes out with `device_id=` instead of `license_key=` → if that device has any guest credits, the transcription succeeds and burns them, and the user is never told their license stopped working. If it has none, they get a confusing `insufficientCredits`/402 rather than `.unauthorized`. Either way the diagnosis the Sentry issue was chasing is now hidden.

(Entitlement itself is still server-enforced, so this is not a paid-tier bypass — but it is a silent identity switch and a wrong user-facing error.)

Gate the retry on the revalidation verdict rather than on whatever `getTranscriptionIdentifier()` happens to return:

```suggestion
            let revalidation = await licenseManager.validateLicense(identifier)
            let (retryIdentifier, retryIsLicensed) = await licenseManager.getTranscriptionIdentifier()
            // Only retry when revalidation confirmed the license is still good and
            // still resolves to a licensed identifier. A confirmed-dead license must
            // surface .unauthorized, not silently re-send as an unlicensed device.
            guard revalidation.isValid, retryIsLicensed else { throw mapped }

            let retryResponse = try await sendTranscribeRequest(identifier: retryIdentifier, isLicensed: retryIsLicensed)
```

ray-amjad:

---

### `app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift:410` — ✅ **CONFIRMED**

**Original claim:** The revalidate-and-retry block is unreachable because RustRetry.perform only returns 2xx, so a 401/403 is thrown before reaching the catch that wraps the parse call — the PR therefore does not fix HYPERWHISPER-T2.

**What the code actually does:** Three independent facts prove it. (1) HyperWhisperCloudProvider.swift:383 calls `var response = try await sendTranscribeRequest(...)` OUTSIDE any do/catch (the surrounding `do` at line 388 wraps only `hyperwhisperCloudParseTranscribeResponse(resp: response)` at line 389); `grep -n 'do {'` shows the nearest enclosing do-blocks are 320-322 (inside the nested func, for request-building) and 388, so a throw from line 383 exits `transcribe(...)` entirely. (2) RustRetry.swift:109-111 returns only when `(200...299).contains(Int(response.status))`; every non-2xx path either retries (line 129-131) or, on give-up, `throw enrichRateLimited(parseError(response), ...)` at line 136 — i.e. it throws a Swift `TranscriptionError`, never an `HwTranscriptionError`, so the typed `catch let err as HwTranscriptionError` at line 390 could not receive it even if it were in scope. (3) Even the error type reaching line 390 can never be Unauthorized: shared-core-rs/crates/hw-net/src/providers/hyperwhisper_cloud.rs:199→204 `parse_routed_response` only calls `classify_status`/`classify_status_json` when `resp.status != 200` (lines 210-222), and hyperwhisper_cloud.rs:308-309 maps ONLY status 401|403 to `TranscriptionError::Unauthorized`; a 2xx body failure yields NoSpeech or Parse. RustRetry.swift:239-240 shows `.unauthorized` is produced solely from `HwTranscriptionError.Unauthorized`. Therefore `mapped` at line 397 can never be `.unauthorized`, the guard at line 410 always throws, and lines 414-429 are dead code.

**Corrected statement:** The revalidation/retry block (lines 410-429) is genuinely unreachable, so the PR does not fix HYPERWHISPER-T2. A real fix must catch the `TranscriptionError.unauthorized` thrown out of `sendTranscribeRequest` at line 383 (e.g. wrap that call in do/catch on TranscriptionError), not the HwTranscriptionError from the 2xx-body parse.


### `app/macos/hyperwhisper/Managers/Transcription/Providers/Cloud/HyperWhisperCloudProvider.swift:418` — 🟡 **PARTLY TRUE**

**Original claim:** Even if reachable, the validateLicense result is discarded and the retry re-sends on the guest/device-ID wallet, silently downgrading a licensed user instead of failing fast.

**What the code actually does:** The mechanism is exactly as described: line 415 is `_ = await licenseManager.validateLicense(identifier)` (return value discarded); line 416 re-reads `licenseManager.getTranscriptionIdentifier()`, which at LicenseManager.swift:191-198 returns `(key, true)` only when `licenseStatus == .active` and otherwise falls through to `return (DeviceIdentifierGenerator.generate(), false)`. `processValidationResult` (LicenseManager.swift:140-146) assigns `licenseStatus = result.status` unconditionally, and LicenseNetworkService.swift:184-215 returns `.invalid` both for a server rejection and for an offline validation whose 7-day grace has lapsed. So a confirmed-revoked (or offline-ungraced) license would make line 418 re-send the same audio with `device_id` (RustCoreMapping.transcribeParams at line 291: `deviceID: isLicensed ? nil : identifier`) — the guest wallet — with no check on `retryIsLicensed`. What is wrong is the stated consequence: this can never happen, because the enclosing branch is dead code (see finding 1), so no user is downgraded today. It is a latent defect in the fix's design, not a shipping behavior.

**Corrected statement:** Accurate as a code-shape critique but not a live defect: the block is unreachable, so no silent guest-wallet downgrade occurs. If the reachability bug is fixed, line 418 must first check `retryIsLicensed` (and/or the discarded `LicenseValidationResult`) and rethrow `.unauthorized` rather than retrying on the device-ID wallet.


---

PR #52, opened 2026-07-17
fix(macos): keep missing Parakeet models local and require download

ray-amjad:
## Problem

Parakeet-backed modes could reach transcription with no compatible model installed. The previous implementation attempted to recover by routing to HyperWhisper Cloud, which created two review-confirmed problems:

- an explicit Local API request for `engine: "parakeet"` could upload audio off-device;
- a locally configured Parakeet mode could silently change its privacy and billing behavior.

## Fix

- Keep Parakeet requests local. Missing models now fail with the normal model-download requirement instead of silently switching providers.
- Canonicalize the default Parakeet V3 model identifier in Local API resolution.
- Check model availability before runtime preparation and handle a model-disappeared race without reporting a misleading Sentry failure.
- Surface a visible orange `Parakeet Vx (Download Required)` status so the user knows how to recover.
- Observe availability for the exact selected Parakeet version and re-prepare it once after the user completes the download, clearing the stale status without polling or automatically starting a download.
- Do not automatically re-download a model the user may have deliberately removed.

## Verification

- Thread-aware review passes covered the original four unresolved review threads, the overlap note with PR #48, and the post-push readiness-recovery follow-up.
- `git diff --check` passed.
- Unsigned macOS Debug `xcodebuild` passed.

Refs: HYPERWHISPER-SV, HYPERWHISPER-TD.


app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
```diff
@@ -718,6 +718,22 @@ class TranscriptionProviderRouter {
...
+            // even though nothing ever implemented the "cloud fallback" this app
+            // logs elsewhere. Check availability up front and actually route to
+            // HyperWhisper Cloud instead of hard-failing the transcription.
+            guard parakeetProvider.isAvailable(for: modelId) else {
+                AppLogger.transcription.warning("⚠️ Parakeet model not downloaded, falling back to HyperWhisper Cloud: \(modelId, privacy: .public)")
+                ensureHyperWhisperCloudProvider()
+                guard let hwProvider = hyperwhisperCloudProvider else {
+                    throw TranscriptionError.modelNotDownloaded
+                }
+                return hwProvider
```
ray-amjad: **Fallback returns the cloud provider without any of the gating the real cloud path applies.**

Every other route to `hyperwhisperCloudProvider` (`selectProvider`, both overloads — see lines ~577-587 and the `.hyperwhisper` arm above) runs three checks after picking the instance:

```swift
if let healthManager = providerHealthManager {
    let status = await healthManager.ensureHealthy(cloudProviderType)
    if !status.isHealthy { throw errorForHealthStatus(status, ...) }
}
guard provider.isAvailable else { ... }
if !NetworkStatus.shared.isOnline { throw TranscriptionError.transientNetwork(...) }
```

This `return hwProvider` skips all of them, and because the local branch `return`s early out of `selectProvider`, they never run later either.

Concrete failure: user's license has expired (or the Fly backend is unreachable). A normal Cloud mode gets a fast, correctly-typed `.unauthorized(provider:)` / `.transientNetwork("Provider unreachable")` toast from `ensureHealthy`. The Parakeet-fallback user instead uploads the whole recording and eats whatever raw server/transport error comes back — different toast, different retry classification, and a full upload's worth of latency before it fails.

PR #48 (`fix/parakeet-fallback-HYPERWHISPER-TD`) solves the same bug with a `selectCloudFallbackProvider()` helper that replays exactly these three checks. Worth adopting that shape here.

app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift
```diff
@@ -183,6 +183,22 @@ class TranscriptionModelManager: ObservableObject {
...
+            // `prepareIfNeeded` unconditionally, which throws `.modelNotDownloaded`
+            // into a hard user-facing error dialog plus an unconditional Sentry
+            // capture on every affected launch. A real transcription attempt now
+            // falls back to HyperWhisper Cloud (`TranscriptionProviderRouter.
+            // selectLocalProvider`), and preloading isn't required for that
+            // fallback to work — so treat "not downloaded" like the cloud/no-mode
+            // case above instead of erroring.
+            guard parakeetProvider.isAvailable(for: modelId) else {
+                AppLogger.models.warning("Parakeet model not downloaded, will use HyperWhisper Cloud fallback at transcribe time: \(modelId, privacy: .public)")
+                modelReadyState = .ready(name: "Cloud")
```
ray-amjad: **The missing model never self-heals, and the user is never told they left local transcription.**

Neither call site kicks off a download or surfaces anything to the user — the only trace is an `AppLogger` warning. `ParakeetModelManager.startDownload(_:)` (`Managers/Transcription/Models/ParakeetModelManager.swift:193`) already exists and is what the Models UI calls.

Concrete scenario: user picks a Parakeet mode, the weights were evicted (or a restored backup points at a model this install never downloaded). Before this PR they got `"Failed to prepare Parakeet"` — noisy, but it told them to go download it. After this PR the status bar reads a confident **"Cloud"**, and *every* subsequent dictation silently uploads audio to HyperWhisper Cloud and bills their credits. Nothing ever re-downloads the model, so this is permanent until the user independently notices and visits Settings → Models.

Two distinct problems fall out of that:

1. **Credits.** A local Parakeet mode is free and unmetered. Silently rerouting it to metered cloud spends the user's balance on transcriptions they believe are on-device; when the balance runs out they get an `insufficientCredits` error from a mode that has no obvious connection to Cloud.
2. **Privacy.** Local Parakeet is a deliberate choice to keep audio on-device. Uploading it to the backend with no notification is a meaningful default to change silently.

PR #48 handles both: it calls `startDownload(modelId)` at each site and shows `"Parakeet V3 isn't downloaded yet — downloading now. Using cloud until it's ready."` That makes the fallback self-healing and one-time rather than silent and permanent. At minimum this PR needs the `startDownload` call plus a user-visible notice.

app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift
```diff
@@ -718,6 +718,22 @@ class TranscriptionProviderRouter {
...
+            // was never downloaded (or was evicted after the fact) — `prepareIfNeeded`
+            // used to throw `.modelNotDownloaded` straight into the caller here,
+            // even though nothing ever implemented the "cloud fallback" this app
+            // logs elsewhere. Check availability up front and actually route to
+            // HyperWhisper Cloud instead of hard-failing the transcription.
+            guard parakeetProvider.isAvailable(for: modelId) else {
+                AppLogger.transcription.warning("⚠️ Parakeet model not downloaded, falling back to HyperWhisper Cloud: \(modelId, privacy: .public)")
+                ensureHyperWhisperCloudProvider()
+                guard let hwProvider = hyperwhisperCloudProvider else {
+                    throw TranscriptionError.modelNotDownloaded
```
ray-amjad: **Wrong error type here, and this PR's own analysis explains why it will be invisible.**

Reaching this line means `ensureHyperWhisperCloudProvider()` could not construct the provider — i.e. `licenseManager` or `creditManager` is nil (line ~589). That is a provider-initialization failure, not a missing model, and every other site that hits it throws `.providerNotAvailable(provider: "HyperWhisper Cloud", reason: "Failed to initialize provider")` (lines ~267 and ~505), with the 267 site also dropping a Sentry breadcrumb recording which manager was nil.

Two consequences of using `.modelNotDownloaded` instead:

1. The user is told to download a model, which will not fix anything — the model genuinely being absent is the *precondition* for reaching this branch, not the failure.
2. Per this PR's own description, `shouldCaptureTranscriptionErrorInSentry` deliberately excludes `.modelNotDownloaded` as "user-recoverable". So a real wiring bug (managers not injected) is swallowed silently — exactly the reporting blind spot the PR body identifies as the reason SV looked stale.

```suggestion
                guard let hwProvider = hyperwhisperCloudProvider else {
                    throw TranscriptionError.providerNotAvailable(provider: "HyperWhisper Cloud", reason: "Failed to initialize provider")
                }
```

ray-amjad:
Review note — **overlap with PR #48** (`fix/parakeet-fallback-HYPERWHISPER-TD`).

Both PRs fix the same two call sites for the same two Sentry issues, and they will conflict textually on merge. They are not just duplicates — they diverge semantically, and #48 is strictly the more complete of the two:

| | #52 (this) | #48 |
|---|---|---|
| Detects missing model | `parakeetProvider.isAvailable(for:)` (disk check) | `parakeetModelManager.isModelInstalled(_:)` + a `catch .modelNotDownloaded` around `prepareIfNeeded` for the eviction race |
| Cloud fallback gating | none — returns the provider raw | `selectCloudFallbackProvider()` replays health / `isAvailable` / online checks |
| Re-downloads the model | no — fallback is permanent | yes, `startDownload(modelId)` |
| Tells the user | no (oslog only) | yes, `onStateChange?(.error(...))` |
| Error when provider init fails | `.modelNotDownloaded` (wrong type; excluded from Sentry) | `.providerNotAvailable` |

Recommendation: merge one, close the other — do not merge both. Taking **#48** and closing this one is the smaller amount of work; taking this one means porting #48's gating helper, `startDownload` call, user notice, and error type across, at which point they are the same PR.

One thing #52 has that #48 does not: it checks availability **before** setting `modelReadyState = .loading(...)`, so the status bar never flashes "Parakeet V3" for a model that is not there. Cheap to carry over to #48.

ray-amjad:

**Result: 3 partly true.**

---

### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift:734` — 🟡 **PARTLY TRUE**

**Original claim:** The Parakeet→cloud fallback returns the cloud provider while bypassing all three gates the real cloud path applies (health check, entitlement, credit check).

**What the code actually does:** The bypass mechanism is real: selectProvider dispatches to selectLocalProvider at line 195 and returns its result directly, so the fallback's `return hwProvider` (line 734) never reaches the post-selection block at lines 299-330. But the named gates do not do what the claim says. (a) Health check — selectProvider:300-306 calls healthManager.ensureHealthy(cloudProviderType), and CloudProviderHealthManager.performHealthCheck:467-471 returns `.healthy` unconditionally for `.hyperwhisper`, so this gate is a no-op for HW Cloud. (b) `provider.isAvailable` gate at selectProvider:309-325 is also a no-op: HyperWhisperCloudProvider.swift:67 is `var isAvailable: Bool { true }`. (c) Entitlement check — grep for `isLicensed|entitle|hasValidLicense|canUseCloud` across TranscriptionProviderRouter.swift and TranscriptionPipeline+Transcription.swift returns zero hits; there is no client-side entitlement gate to bypass (entitlement is server-side, per repo policy). (d) Credit check — no credit gate exists in selectProvider either; credits are enforced by the server, with HyperWhisperCloudProvider.swift:348-358 mapping a 402 response to TranscriptionError.insufficientCredits inside transcribe(). The one real check that IS skipped is the offline guard at selectProvider:327-330 (`NetworkStatus.shared.isOnline` → .transientNetwork).

**Corrected statement:** The fallback does return early and therefore skips the post-selection block at TranscriptionProviderRouter.swift:299-330 — but two of the three named gates don't exist (no client-side entitlement or credit check anywhere in the router; credits are enforced server-side via 402 in HyperWhisperCloudProvider.transcribe), and the health check it does skip is a hardcoded `.healthy` for `.hyperwhisper`. The only substantive skipped check is the `NetworkStatus.shared.isOnline` guard at line 327-330, so an offline user with a missing Parakeet model gets a raw network failure from inside transcribe() instead of the clean 'No internet connection' error. Secondary real concern (not claimed): the fallback hardcodes HyperWhisper Cloud, ignoring the mode's configured `cloudProvider`.


### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionModelManager.swift:198` — 🟡 **PARTLY TRUE**

**Original claim:** The missing model never self-heals and the user is never told they silently left local transcription; only an AppLogger warning is emitted.

**What the code actually does:** The 'only an AppLogger warning' half is false: line 198 sets `modelReadyState = .ready(name: "Cloud")`, a @Published property (TranscriptionModelManager.swift:77) consumed by ModelStatusBar.swift:56 (`.onReceive(...$modelReadyState)`), which at ModelStatusBar.swift:100-105 renders a BLUE dot plus the literal text "Cloud" instead of the green dot + "Parakeet V3" — the same visible treatment as an explicitly-chosen cloud mode (line 148). The 'never self-heals' half is also false in the recovery sense: prepareModel is re-invoked on every non-cloud recording start (RecordingTranscriptionFlow+StartRecording.swift:345-353) and on every mode change (AppState.swift:836-846), so once the model exists on disk the `parakeetProvider.isAvailable(for: modelId)` guard at line 196 passes and normal local preparation resumes at lines 202-209 with no user action. What is genuinely true: nothing triggers or offers a download of the missing model.

**Corrected statement:** The state is surfaced to the user (ModelStatusBar shows a blue 'Cloud' indicator instead of the Parakeet model name) and recovers automatically on the next mode change or recording start once the model is present. The accurate residual gap is narrower: neither this code nor the pre-PR behaviour auto-downloads or prompts to download the missing Parakeet model, and there is no one-time toast explaining the substitution — only the persistent status-bar indicator.


### `app/macos/hyperwhisper/Managers/Transcription/Coordinators/TranscriptionProviderRouter.swift:732` — 🟡 **PARTLY TRUE**

**Original claim:** Throws .modelNotDownloaded when the real cause is provider-init failure (nil licenseManager/creditManager); wrong error type, and it is invisible to the user.

**What the code actually does:** The error-type inconsistency is real: line 731-733 throws `.modelNotDownloaded` when `ensureHyperWhisperCloudProvider()` (lines 588-597) leaves `hyperwhisperCloudProvider` nil because the weak `licenseManager` (line 80) or `creditManager` (line 83) is nil, whereas the primary cloud path handles the identical nil-provider condition at lines 254-267 by adding a Sentry breadcrumb and throwing `.providerNotAvailable(provider: "HyperWhisper Cloud", reason: "Failed to initialize provider")`. But 'invisible to the user' is refuted: TranscriptionError.swift:52-53 maps `.modelNotDownloaded` to the localized string `transcription.error.modelNotDownloaded` ('Please download a model first') with guidance at lines 180-181; the throw propagates out of selectProvider to the catch in TranscriptionPipeline+Transcription.swift:374-378, which sets `state = .error(message: error.localizedDescription)`, and AppState.swift:373-375 turns that state into `showError(message)` — a visible error pill. The comment at TranscriptionPipeline+ErrorClassification.swift:154-164 explicitly documents that `.modelNotDownloaded` is deliberately not Sentry-captured precisely because 'the app already surfaces a user-facing error'.

**Corrected statement:** The user DOES see a localized error ('Please download a model first') via TranscriptionPipeline+Transcription.swift:378 → AppState.showError. The valid part of the finding is the inconsistency/diagnostics gap: this branch should mirror lines 254-267 (Sentry breadcrumb + `.providerNotAvailable(provider:reason:)`) so a nil licenseManager/creditManager is distinguishable from a genuinely absent model. Note this is a rare defensive path — the managers are weak refs that are non-nil whenever the app holds them — and the message shown is arguably still correct guidance in this scenario, so severity is low/nit, not a user-facing bug.


---

PR #54, opened 2026-07-17
Add lightweight macOS PR build+test CI gate

ray-amjad:
## Why

PRs touching the macOS app (Swift) currently have no compile/test signal — reviews are code-review-only. This came up directly while working HYPERWHISPER-F4 (PR #50): the review sandbox has no macOS/Xcode toolchain, so a Swift-only fix could only be verified by code review, not by actually building it.

## What

New `.github/workflows/macos-ci.yml`, triggered on `pull_request`/`push` to `main` (path-filtered to `app/macos/**` and `shared-core-rs/**`), that:

- selects Xcode the same way `macos-release.yml` already does (pin 26.3, fall back sanely)
- runs `xcodebuild build-for-testing` + `test-without-building` against the `hyperwhisper` scheme, Debug config, with signing fully disabled (`CODE_SIGNING_ALLOWED=NO` etc — no keychain/cert/provisioning needed)
- scoped to the `hyperwhisperTests` unit test target only — `hyperwhisperUITests` needs Accessibility/Input-Monitoring permissions that can't be granted non-interactively on a hosted runner, so it's excluded to avoid a flaky/hanging gate
- trusts the committed prebuilt `libhyperwhisper_core.a` (Xcode has no build phase that rebuilds it from `shared-core-rs/`; Rust itself is already covered separately by `shared-core-tests.yml`)
- lets `xcodebuild`'s native exit code propagate — no `xcpretty` swallowing, so build/test failures actually fail the check

## Explicitly NOT included

No archiving, signing, notarization, DMG creation, or publishing — that's `macos-release.yml`'s job. This is purely a build+test gate, deliberately not a trimmed-down copy of the release workflow.

## Verification

No local Xcode/macOS toolchain available to test this workflow before pushing; the real test is whether it actually triggers and passes as a GitHub Actions check once this PR is open (path filters mean it should fire on this diff since it touches its own workflow file). Watching that live run's outcome next.

---

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -1,0 +1,101 @@
+# Lightweight PR gate for the macOS app: build + unit test, no signing.
+#
+# This is deliberately NOT a trimmed-down macos-release.yml — it skips
+# archiving, code signing, notarization, DMG creation, and publishing
+# entirely. Its only job is to prove `main` and open PRs still compile and
+# pass unit tests on a real macOS/Xcode toolchain.
+name: macOS CI
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "app/macos/**"
+      - "shared-core-rs/**"
+      - ".github/workflows/macos-ci.yml"
+  push:
+    branches: [main]
+    paths:
+      - "app/macos/**"
+      - "shared-core-rs/**"
+      - ".github/workflows/macos-ci.yml"
+
+concurrency:
+  group: macos-ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-and-test:
+    name: Build & Test (macOS, unsigned)
+    runs-on: macos-26
+    steps:
+      # ----------------------------------------------------------------
+      # 1. Checkout
+      # ----------------------------------------------------------------
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # ----------------------------------------------------------------
+      # 2. Select Xcode version
+      # ----------------------------------------------------------------
+      # Mirrors macos-release.yml's selection exactly: pin to Xcode 26.3
+      # because 26.4 has strict concurrency errors in FluidAudio. Debug and
+      # release builds must compile against the same toolchain.
+      - name: Select Xcode 26.3
+        run: |
+          XCODE_PATH=$(ls -d /Applications/Xcode_26.3*.app 2>/dev/null | sort -V | tail -1)
+          if [ -z "$XCODE_PATH" ]; then
+            XCODE_PATH=$(ls -d /Applications/Xcode_26*.app 2>/dev/null | grep -v -i beta | sort -V | tail -1)
+          fi
+          if [ -z "$XCODE_PATH" ]; then
+            XCODE_PATH=$(ls -d /Applications/Xcode*.app 2>/dev/null | sort -V | tail -1)
+          fi
+          echo "Using Xcode at: $XCODE_PATH"
+          sudo xcode-select -s "$XCODE_PATH"
+          xcodebuild -version
+
+      # ----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ray-amjad:
## Review round 1 — Claude (medium tier) + Codex

**Codex's P1 (actions/checkout blocked by `allowed_actions: local_only`) — investigated and REFUTED.** `shared-core-tests.yml`'s comment claims marketplace actions startup-fail on this repo, but I pulled the actual GitHub Actions run log for the most recent `macos-release.yml` run (`gh run view 28564586028 --log-failed`) and `actions/checkout@v4` + `actions/setup-node@v4` both downloaded and executed successfully — the run failed later, unrelated to checkout. So `shared-core-tests.yml`'s comment is stale/incorrect for this repo today; not acting on Codex's P1, but flagging that comment as worth a cleanup separately.

**Codex's P2 = Claude's #1 (both agree, high confidence): path filters miss bundled shared resources.** `shared-app-classification/`, `shared-models/`, and `shared-prompts/` are bundled into the Xcode target and read by tests (`CatalogLanguageParityTests`, `AppTypeClassifierTests`, etc.) but aren't in the `paths:` filter — a catalog-only PR would merge with zero build/test signal.

**Additional Claude findings:**

2. `concurrency.group` uses `github.ref`, which is the same literal (`refs/heads/main`) for every push to main — `cancel-in-progress: true` means a later main push cancels CI for an earlier one before it reports, losing that commit's pass/fail record. Should scope cancellation to PR runs only (or key the group on the commit SHA for push events).
3. `build-for-testing` still compiles `hyperwhisperUITests` even though the step's own comment says it's excluded — `-only-testing:hyperwhisperTests` only filters what *runs* in the later `test-without-building` step, not what gets built. Any UI-test-only compile error still fails this "unit-test-only" gate.
4. Xcode-selection shell script is duplicated verbatim from `macos-release.yml` — a future version bump has to be made in both places or they silently diverge.
5. The 4-line code-signing-disable flags are repeated on both `build-for-testing` and `test-without-building` instead of set once (e.g. job-level `env:`).
6. No SwiftPM/DerivedData caching — every run re-resolves/re-clones 9+ packages from scratch (efficiency, not correctness).

Fixing 1, 2, 3, 5 (and 4/6 if straightforward without overengineering) now.

ray-amjad:
## Review round 2 — Claude (medium tier) + Codex

**Codex's P1 — CONFIRMED via live evidence, this is real and currently breaking the CI gate.** Round 1's fix moved the 4 code-signing-disable flags (`CODE_SIGNING_ALLOWED=NO` etc.) from per-invocation `xcodebuild` arguments into a job-level `env:` block. I pulled the actual GitHub Actions run history for this branch:

- commit `f2a5fde` (before round 1) → `macOS CI` run **succeeded**
- commit `3a36a13` (after round 1's env-block consolidation) → `macOS CI` run **failed**, with `error: Signing for "hyperwhisper" requires a development team. Select a development team in the Signing & Capabilities editor that matches the selected profile "HyperWhisper Developer ID".`

So job-level `env:` vars do NOT override this project's explicit Manual signing settings the way passing them as `xcodebuild` command-line arguments does — Claude's own reviewer this round reasoned this was "the standard CI idiom" and refuted the concern, but that reasoning was wrong for this specific project's signing configuration; the live run is the actual proof. Revert to passing all 4 flags directly on both `xcodebuild` invocations (as the original round-0 version did), not via job-level `env:`.

**Claude findings (both minor, keep if cheap):**
1. The Xcode-version 3-tier fallback silently substitutes a different Xcode version with no warning if the pinned 26.3 is missing — could silently land on the known-broken-for-FluidAudio 26.4. Pre-existing pattern copied from `macos-release.yml`, but now duplicated into a second workflow.
2. `paths:` filter list is duplicated verbatim between the `pull_request` and `push` triggers with no shared source — minor future-drift risk.

Fixing the signing regression now (this is the priority — the gate is currently red). Will also address the two minor items if straightforward.

PR #61, opened 2026-07-28
fix(windows): dispose WhisperProcessor asynchronously so cancels don't surface as failures

ray-amjad:
## Problem

User report (Windows 1.8.0):

> Transcription failed: Cannot dispose while processing, please use DisposeAsync instead.

## Root cause

`TranscriptionService.cs` built the Whisper processor with a synchronous `using`:

```csharp
using var processor = builder.Build();
await foreach (var _ in processor.ProcessAsync(audioStream, cancellationToken)...)
```

Whisper.net 1.9.0's `WhisperProcessor.Dispose()` opens with a hard guard:

```csharp
if (processingSemaphore.CurrentCount == 0)
    throw new Exception("Cannot dispose while processing, please use DisposeAsync instead.");
```

That semaphore is held by the background task running native `whisper_full`. When the cancellation token trips, `ProcessAsync` throws `TaskCanceledException` immediately — it does **not** wait for that native task to unwind. The `using` then calls the sync `Dispose()`, the semaphore is still held, and the raw `Exception` replaces the cancellation.

Downstream, `MainViewModel` has a clean `catch (OperationCanceledException)` handler that would have shown "Recording cancelled". The swapped exception falls through to the generic `catch (Exception ex)` instead, producing `errors.transcriptionFailed` with the dispose message and marking the transcript Failed in History.

Secondary damage: `Dispose()` threw before freeing the processor's pinned `GCHandle`s and unmanaged strings, so every cancelled local transcription leaked them.

Triggers: Escape while transcribing, the file-transcription progress Cancel button, and app shutdown mid-transcription.

## Fix

- `await using var processor` — `DisposeAsync()` waits on the semaphore first. The native call is already aborting via Whisper.net's `OnAbort` callback (wired to the same token), so it returns promptly rather than blocking for the full inference.
- Added `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)` in `TranscribeFileInternalAsync` so cancels log at Info instead of Error and propagate as cancellations.
- Routed the sync `TranscribeFile()` wrapper through `TranscribeFileAsync` (already `Task.Run`-wrapped) so the new `DisposeAsync` continuation can't deadlock if a caller blocks on it from the UI thread.

## Verification

`dotnet build -c Debug` compiles clean — `HyperWhisper -> bin/Debug/net10.0-windows10.0.19041.0/HyperWhisper.dll`. The build then stops on the pre-existing `hyperwhisper_core.dll (x64) is missing from Resources\rust-core\x64\` packaging check, which is unrelated to this change and expected on a machine without the Rust core built.

Not runtime-tested on Windows — the analysis is source-level against Whisper.net 1.9.0 plus the compile check. Worth a manual pass: start a local Whisper transcription, hit Escape mid-flight, confirm the overlay shows "Recording cancelled" and no Failed row lands in History.

## Note

Sentry has no matching issue (searched `ray-amjad/hyperwhisper`, 90d) — this exception is caught and toasted, never captured, so there's no telemetry on blast radius.

https://claude.ai/code/session_01W52sLtKvaSwcvrLYyb8bJ7

### DIFF
--- a/app/windows/HyperWhisper/Services/TranscriptionService.cs
+++ b/app/windows/HyperWhisper/Services/TranscriptionService.cs
@@ -407,9 +407,12 @@ public class TranscriptionService : ITranscriptionProvider, IDisposable
     /// <returns>Transcribed text.</returns>
     public string TranscribeFile(string audioPath, string? language = null)
     {
-        // Go through TranscribeFileAsync so the body runs on a thread-pool thread with
-        // no SynchronizationContext — blocking on it from the UI thread would otherwise
-        // risk deadlocking on the processor's `await using` DisposeAsync continuation.
+        // Go through TranscribeFileAsync so the body runs on a thread-pool thread with no
+        // SynchronizationContext. Not fixing a live deadlock — today's first suspension
+        // point already uses ConfigureAwait(false), so nothing captures the UI context.
+        // It's belt-and-braces: neither `await using` nor Whisper.net's DisposeAsync uses
+        // ConfigureAwait(false), so a future edit that reaches the dispose with a captured
+        // context would deadlock any UI-thread caller blocking on this wrapper.
         return TranscribeFileAsync(audioPath, language)
             .GetAwaiter()
             .GetResult();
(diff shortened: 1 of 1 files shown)


ray-amjad:
## Follow-up from code review (`fdc676d`)

The review caught that the first commit only fixed half the problem. Confirmed and fixed.

**Important correction to the original PR description:** there is no `v1.9.0` tag in the whisper.net repo, and the `main` source I originally read differs materially from the pinned 1.9.0 assembly. Decompiling `~/.nuget/packages/whisper.net/1.9.0/lib/net8.0/Whisper.net.dll` shows 1.9.0's `ProcessInternalAsync` has **no** cancellation check around `throw new WhisperProcessingException(num)` — that check only exists in later versions:

```csharp
// 1.9.0, decompiled
int num = nativeWhisper.Whisper_Full_With_State(...);
if (num != 0)
{
    throw new WhisperProcessingException(num);   // no ThrowTaskCanceledIfCancellationRequested
}
```

Whisper.net wires its abort callback to our token (`WhisperAbortEventHandler` / `currentCancellationToken`), so cancelling makes `whisper_full` return non-zero and the background task fault with `WhisperProcessingException`. In `ProcessAsync`'s state machine:

```csharp
IL_026c:
if (!processingTask.IsCompleted || !buffer.IsEmpty) {
    cancellationToken.ThrowIfCancellationRequested();   // skipped when task already faulted
    ...
}
awaiter = processingTask.ConfigureAwait(false).GetAwaiter();
IL_02fd:
awaiter.GetResult();                                    // rethrows WhisperProcessingException here
if (cancellationToken.IsCancellationRequested) throw new TaskCanceledException();  // never reached
```

Once the task has faulted and the segment buffer is empty, the guard is false, so the cancellation check is skipped and `GetResult()` rethrows `WhisperProcessingException` before the `TaskCanceledException` line.

So **which exception surfaces is a race**: cancelling at a segment boundary gives `OperationCanceledException`, cancelling during the encode pass gives `WhisperProcessingException`. My narrow `catch (OperationCanceledException)` filter only covered the first — Escape mid-encode still produced a "Transcription failed" toast, just with a different message.

**Fix:** broaden the filter to any exception observed while the token is cancelled, and normalize non-OCE faults into `OperationCanceledException` (original preserved as `InnerException`, type name logged) so callers' cancel handlers run either way.

Also corrected the `TranscribeFile()` comment — routing through `TranscribeFileAsync` is belt-and-braces, not a fix for a live deadlock, since the first suspension point already uses `ConfigureAwait(false)`.

### Known limitation, not addressed

`DisposeAsync()` does `await processingSemaphore.WaitAsync()` with no timeout and no token, so cancel latency is now bounded by how fast whisper.cpp polls its abort callback rather than being instant. In the pathological case where a native pass stops polling, this would hang — and since `_inFlight` is only decremented in the outer `finally`, the drain loops in `InitializeAsync` and `UnloadModelAsync` would spin too, hanging a model switch. I left it: a bounded wait would mean abandoning dispose while native code is still running, which reintroduces the use-after-free this PR exists to avoid. Worth revisiting if it ever shows up in the wild.

PR #79, opened 2026-07-30
fix(windows): stop history from destroying completed transcriptions

### LINKED ISSUE #75: Windows: completed transcriptions are overwritten with "Transcription did not finish" while History is open
# Windows: completed transcriptions are overwritten with "Transcription did not finish" while History is open

**Version:** 1.8.2 (x64), Windows 11 26200

## Symptom

Recordings transcribe correctly and the text pastes into the target app, but the History row then flips to **Failed / "Transcription did not finish"**. The text is gone from the database — `Text`, `TranscribedText`, `PostProcessedText` and `TranscriptionProvider` are all wiped, so the only recovery is Retry Transcription. A related variant leaves the row showing "Processing" indefinitely even though the transcription completed.

My database currently holds **425** rows with `FailedReason = "Transcription did not finish"`, dated 2026-06-30 through today — every Failed row I have except one genuine "insufficient credits".

## The trigger

**It only happens while the History page is open.** From one day's log:

```
15:05:52  APPLICATION STARTING
          ... ~2 hours of normal use, zero failures ...
17:11:43  HistoryPage: Loaded
17:12:27  first clobber (the very next recording)
17:12:57  HistoryPage: Unloaded
...
21:27:40  HistoryPage: Unloaded
          ... 26 minutes, zero failures ...
21:53:24  HistoryPage: Loaded
21:55:10  clobber
21:55:48  clobber
```

24 clobbers that day, every one with History loaded, none otherwise. `HistoryViewModel` only exists after the page is first navigated to, which is why it looks like something degrading over time.

## What the log shows

A single transcript, one flow, one thread — saved correctly, then destroyed 400 ms later:

```
21:22:48.921  FILE TRANSCRIPTION COMPLETE   Segments: 19   Characters: 1031
21:22:51.681  Post-processing complete (1031 -> 1042 chars)
21:22:51.753  SmartPasteService: Sent Ctrl+V to paste text
21:22:51.756  HistoryService: Updated transcript 30e3de90 (Status: Completed)
21:22:51.757  [PERF] TranscriptionFlow: Paste done, history save attempted
21:22:51.757  RecordingOverlayWindow: Showing success state
21:22:52.164  RecordingOverlayWindow: Hidden
21:22:52.166  HistoryService: Updated transcript 30e3de90 (Status: Failed)
21:22:52.167  EnsureTranscriptTerminalStatus: Flipped orphaned transcript 30e3de90 from Processing to Failed
```

No `ERROR` entries anywhere in the file — nothing actually failed.

## Root cause

`HistoryViewModel.OnTranscriptUpdated` obtains the entity through `vm.ToEntity()`, but `ToEntity()` is a **save-back**: it writes the view model's own snapshot *into* the wrapped entity (`TranscriptViewModel.cs:143`).

The entity it wraps is normally the same instance the event is carrying. `MainViewModel` passes its live `Transcript` to `HistoryService.UpdateTranscript`, which forwards that instance to subscribers, and `OnTranscriptAdded` wrapped that same instance when the Processing row appeared. So `entity` and `transcript` are aliases:

```csharp
var entity = vm.ToEntity();          // reverts the entity to the snapshot taken
                                     // at construction: Processing, no text
entity.Text = transcript.Text;       // entity IS transcript -> self-assignment
entity.Status = transcript.Status;   // ditto, and so on for every field
vm.Refresh();                        // re-reads the reverted entity
```

Every "refresh" copy assigns an already-reverted value to itself. Two consequences:

1. `vm.Refresh()` re-reads the reverted entity, so the row keeps showing **Processing**.
2. `MainViewModel`'s `finally` block then calls `EnsureTranscriptTerminalStatus`, which reads the same reverted instance, sees `Processing`, and writes **"Transcription did not finish"** over the row that was already saved as Completed — nulling the text fields via `CurrentValues.SetValues`.

The reverting callback is queued with `Dispatcher.BeginInvoke`, and the flow's 400 ms success-overlay `await` is exactly the window in which it runs — just before the `finally`. Whether it lands before or after that block is what decides between the "Failed" variant and the "stuck on Processing" variant.

`EnsureTranscriptTerminalStatus` also contributes: it decides what to write from the in-memory status alone, so it is capable of downgrading a row the database already recorded as Completed.

## Reproduction

1. Open HyperWhisper and navigate to **History**. Leave it open.
2. Record any dictation with any mode and let it complete.
3. The text pastes correctly, but the History row shows Failed / "Transcription did not finish" (or stays on "Processing").

Staying on the Home tab avoids it entirely.

## Fix

PR #<n> — two commits:

1. Replace the `ToEntity()` round-trip in `OnTranscriptUpdated` with a new `TranscriptViewModel.ApplyUpdate`, which copies the incoming entity onto the source only when the two are distinct instances, then refreshes. Same field set as before.
2. Make `EnsureTranscriptTerminalStatus` treat the persisted row as the authority: if the database already holds a terminal status, return without writing. It exists to rescue rows that never finished, not to overrule ones that did.

Plus a regression check in `HyperWhisper.SmokeTests`.

phil-cheesman:
Fixes #75

Completed transcriptions were being overwritten in History with **"Transcription did not finish"**, losing `Text`, `TranscribedText`, `PostProcessedText` and `TranscriptionProvider`. The transcription itself succeeded and the text pasted correctly; the row was destroyed about 400 ms later. A related variant left the row showing "Processing" forever. My database has 425 rows in this state, dated across the past month.

Only reproducible with the History page open — that's when `HistoryViewModel` exists to hold the entity.

### Cause

`HistoryViewModel.OnTranscriptUpdated` read the entity via `vm.ToEntity()`, but `ToEntity()` is a save-back: it writes the view model's own snapshot *into* the wrapped entity. And the entity it wraps is normally the same instance the event carries — `MainViewModel` passes its live `Transcript` to `HistoryService.UpdateTranscript`, which forwards that instance to subscribers, and `OnTranscriptAdded` wrapped it when the Processing row first appeared. So `entity` and `transcript` are aliases:

```csharp
var entity = vm.ToEntity();          // reverts the entity to the snapshot taken at
                                     // construction: Processing, no text
entity.Text = transcript.Text;       // entity IS transcript -> self-assignment
entity.Status = transcript.Status;   // ditto, for every field
vm.Refresh();                        // re-reads the reverted entity
```

Two consequences. `vm.Refresh()` re-reads the reverted entity, so the row keeps showing Processing. And the transcription flow's `finally`-block safety net (`EnsureTranscriptTerminalStatus`) reads the same reverted instance, sees `Processing`, and writes the failure over the row already saved as Completed.

The reverting callback is queued with `Dispatcher.BeginInvoke`; the flow's 400 ms success-overlay `await` is the window in which it runs, immediately before the `finally`. Whether it lands before or after that block is what decides between the "Failed" and "stuck on Processing" variants.

### Changes

Two commits, reviewable separately.

**1. Remove the cause.** New `TranscriptViewModel.ApplyUpdate(Transcript)` copies the incoming entity onto the source only when the two are distinct instances, then refreshes — so the aliased case is a no-op instead of a revert, and a re-read entity still lands. Same field set the handler used before. `ToEntity()` now documents that it is a write, not a getter.

**2. Make the safety net incapable of this.** `EnsureTranscriptTerminalStatus` decided what to write from the in-memory status alone, so it could downgrade a row the database already recorded as Completed. It now treats the persisted row as the authority: if it already holds a terminal status, return without writing. The repair path (persisted Processing, in-memory terminal) and the flip path (both Processing) are unchanged.

The second commit alone would have prevented all 425 destroyed rows, but the first is the actual bug — without it, History still displays stale Processing rows.

### Tests

Adds a check to `HyperWhisper.SmokeTests` that reproduces the aliased path: wrap a Processing transcript, complete that same instance, call `ApplyUpdate`, and assert the entity was not reverted and the view model reflects the result. Also covers the distinct-instance path.

### Verifying

1. Open History and leave it open.
2. Record any dictation and let it complete.
3. The row should show the transcribed text, not "Transcription did not finish" and not a stuck "Processing".
4. Log should contain `HistoryService: Updated transcript … (Status: Completed)` with no following `EnsureTranscriptTerminalStatus: Flipped orphaned transcript …` for the same id.

### Not covered here

Existing rows already destroyed are unrecoverable from the text side — the audio is retained, so Retry Transcription still works on them. Let me know if you'd want a one-time migration that re-marks them, though I'd guess not.

### DIFF
--- a/app/windows/HyperWhisper.SmokeTests/Program.cs
+++ b/app/windows/HyperWhisper.SmokeTests/Program.cs
@@ -25,6 +25,7 @@ using HyperWhisper.Data.Entities;
 using HyperWhisper.Models;
 using HyperWhisper.Services;
 using HyperWhisper.Services.Transcription;
+using HyperWhisper.ViewModels;
 using HyperWhisper.Views.Pages.Settings;
 using uniffi.hyperwhisper_core;
 
@@ -364,6 +365,62 @@ internal static class Program
                 Assert(
                     !AssemblyAIService.IsSyncEligible(Result<double>.Success(cap - 1), cap, isMedicalModel: true),
                     "a medical model should NOT be sync-eligible even with an otherwise-eligible duration — sync has no medical/domain concept");
+            });
+
+            Run("TranscriptViewModel.ApplyUpdate never reverts the entity it wraps", () =>
+            {
+                // Reproduces the History clobber: MainViewModel creates a Processing
+                // transcript, HistoryViewModel wraps that exact instance, then
+                // MainViewModel completes the SAME instance and hands it back through
+                // HistoryService.TranscriptUpdated. Absorbing the update must not
+                // revert the entity to the snapshot taken at construction — the
+                // transcription flow's finally-block safety net reads that status and
+                // would overwrite a completed transcript with a failure.
+                var transcript = new Transcript
+                {
+                    Id = Guid.NewGuid(),
+                    Status = TranscriptStatus.Processing,
+                    Text = "Processing audio..."
+                };
+
+                var vm = new TranscriptViewModel(transcript);
+
+                transcript.Text = "the real transcription";
+                transcript.TranscribedText = "the real transcription";
+                transcript.Status = TranscriptStatus.Completed;
+                transcript.TranscriptionProvider = "Whisper large-v3-turbo";
+
+                vm.ApplyUpdate(transcript);
+
+                Assert(transcript.Status == TranscriptStatus.Completed,
+                    $"entity status was reverted to {transcript.Status}");
+                Assert(transcript.Text == "the real transcription",
+                    $"entity text was reverted to '{transcript.Text}'");
+                Assert(transcript.TranscribedText == "the real transcription",
+                    "entity raw text was discarded");
+                Assert(transcript.TranscriptionProvider == "Whisper large-v3-turbo",
+                    "entity provider was discarded");
+
+                Assert(vm.Status == TranscriptStatus.Completed,
+                    $"view model still shows {vm.Status}");
+                Assert(vm.Text == "the real transcription",
+                    $"view model still shows '{vm.Text}'");
+
+                // A distinct instance (e.g. re-read from the DB) must still land.
+                var reread = new Transcript
+                {
+                    Id = transcript.Id,
+                    Status = TranscriptStatus.Failed,
+                    Text = "No speech detected",
+                    FailedReason = "No speech detected",
+                    RetryCount = 2
+                };
+
+                vm.ApplyUpdate(reread);
+
+                Assert(vm.Status == TranscriptStatus.Failed, $"view model shows {vm.Status}");
+                Assert(vm.Text == "No speech detected", $"view model shows '{vm.Text}'");
+                Assert(vm.RetryCount == 2, $"view model shows retry count {vm.RetryCount}");
             });
 
             Run("BackupExportSettingsPage initializes under WPF", () =>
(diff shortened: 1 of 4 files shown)


ray-amjad:
### Code review

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #81, opened 2026-07-30
fix(windows, macos): keep vocabulary replacements when AI post-processing is off

### LINKED ISSUE #80: Vocabulary replacements are silently dropped when AI post-processing is off (Windows + macOS)
# Vocabulary replacements are silently dropped when AI post-processing is off (Windows + macOS)

**Version:** Windows 1.8.2, and the same shape is present on macOS at `main`.

## Symptom

Switching a mode's AI post-processing off discards **every** vocabulary replacement. My 22 configured corrections (`Purcell` → `Vercel`, `cloud code` → `Claude Code`, and so on) simply stop applying. Nothing in the UI suggests vocabulary is coupled to post-processing, so this reads as vocabulary being broken rather than as a side effect of a different setting.

It matters more than it looks: turning post-processing off is the main way to get a fast fully-local mode, since local LLM post-processing is CPU-only on non-NVIDIA hardware (#77). So the configuration you'd reach for to make local dictation usable is exactly the one that loses your spelling corrections.

## Cause

`app/windows/HyperWhisper/Services/Transcription/TranscriptionOrchestrator.cs:242`:

```csharp
if (applyPostProcessing && mode.PostProcessingMode != 0)
{
    var postProcessingResult = await _postProcessingService.ProcessPreservingBreaksAsync(...);
    // Apply vocabulary replacements whether or not AI post-processing succeeded.
    // Matches macOS + v1.6: when post-processing fails/skips, ProcessAsync returns the
    // original transcription, and the user's vocabulary corrections must still be applied.
    finalText = _vocabularyProcessor.ApplyReplacements(postProcessingResult.Text);
    ...
}
else if (applyPostProcessing)
{
    // No AI post-processing — optionally apply lightweight filler word removal.
    if (SettingsService.Instance.RemoveFillerWords)
    {
        finalText = SmartSpacing.RemoveFillerWords(finalText);
    }
    finalText = TranscriptionTextProcessing.ProcessVoiceCommands(finalText);
    // <- ApplyReplacements never runs here
}
```

The comment describes the intended invariant, but it only covers post-processing *failing or being skipped at runtime* — in which case `ProcessPreservingBreaksAsync` returns the original text and the call still happens. A mode with post-processing **switched off** takes the `else if` branch entirely and never reaches it.

## macOS has the same bug

`app/macos/hyperwhisper/Managers/Transcription/Pipeline/TranscriptionPipeline+Transcription.swift:253`:

```swift
} else {
    // Branch: no post-processing.
    ...
    let withoutFillers = settingsManager?.removeFillerWords == false
        ? text
        : TranscriptionTextProcessing.removeFillerWords(text, language: detectedLanguage ?? languageArg)
    finalText = TranscriptionTextProcessing.processVoiceCommands(withoutFillers)
    // <- applyVocabularyReplacements never runs here either
}
```

Both other branches (`hyperwhisperCloudAIText`, `shouldRunPostProcessing`) call `vocabularyProcessor.applyVocabularyReplacements`; only the no-post-processing branch skips it. So the platforms are consistent with each other — consistently wrong.

## Reproduction

1. Add a vocabulary entry with a replacement, e.g. word `Purcell` → replacement `Vercel`.
2. Use a mode with AI post-processing **enabled**, dictate "deploy to Purcell". Output: "deploy to Vercel".
3. Switch that mode's post-processing to **off**, dictate the same thing. Output: "deploy to Purcell".

## Fix

PR #<n> moves the call into the no-AI branch on Windows, applied last so it mirrors the post-processing branch where it is the final transformation.

Not hoisted above the `if`/`else` on purpose — the post-processing branch would then apply replacements twice, and pairs aren't guaranteed idempotent when one entry's output is another entry's input.

I've left the macOS hunk out of the PR because I can't build or run macOS to verify it. It's the same one-line addition at `TranscriptionPipeline+Transcription.swift:266`. Happy to add it to the same PR if you'd rather have both together, or you may prefer to take that side yourself given `macos-ci.yml` can actually exercise it.

phil-cheesman:
Fixes #80

Turning AI post-processing off on a mode silently discarded every vocabulary replacement the user had configured. Both platforms have the same gap, and both are fixed here — one commit each, reviewable separately.

## Windows

`TranscriptionOrchestrator` applies replacements inside the post-processing branch only:

```csharp
if (applyPostProcessing && mode.PostProcessingMode != 0)
{
    ...
    // "Apply vocabulary replacements whether or not AI post-processing succeeded."
    finalText = _vocabularyProcessor.ApplyReplacements(postProcessingResult.Text);
}
else if (applyPostProcessing)
{
    // filler words + voice commands only — replacements never run
}
```

The existing comment states the intent the code doesn't quite implement. It covers post-processing *failing or being skipped at runtime* — there `ProcessPreservingBreaksAsync` returns the original text and the call still happens. A mode with post-processing **switched off** takes the `else if` branch and never reaches it.

Vocabulary replacements are the user's own spelling corrections, independent of any AI step, so this applies them in the no-AI branch too — last, mirroring the post-processing branch where they're the final transformation.

Deliberately **not** hoisted above the `if`/`else`: the post-processing branch would then apply them twice, and replacement pairs aren't guaranteed idempotent when one entry's output is another entry's input.

Streaming (`applyPostProcessing` false) is untouched — it applies replacements on its own path.

## macOS

Same shape, confirmed at `TranscriptionPipeline+Transcription.swift:224-267` — three branches, and only two of them apply vocabulary:

| Branch | Applies vocabulary? |
|---|---|
| `hyperwhisperCloudAIText` — server-side AI already applied (line 227) | yes |
| `shouldRunPostProcessing` — client-side AI post-processing (line 252) | yes |
| `else` — no post-processing (lines 253-267) | **no** |

The no-AI branch ended at `TranscriptionTextProcessing.processVoiceCommands(withoutFillers)`. Those two were the only `applyVocabularyReplacements` call sites in the app, so with post-processing off nothing applied the user's corrections at all.

Fixed the same way as the Windows side: replacements run last in that branch, after `processVoiceCommands`, and are not hoisted out of the `if`/`else` for the same double-application reason.

Streaming is likewise untouched on macOS — `RecordingTranscriptionFlow.applyStreamingVocabulary` is its own path, and `VocabularyProcessor.applyHardenedReplacement` is the shared per-word logic behind both, so batch and streaming stay consistent.

### Verifying

1. Add a vocabulary entry with a replacement, e.g. `Purcell` → `Vercel`.
2. Mode with post-processing **on**, dictate "deploy to Purcell" → "deploy to Vercel".
3. Same mode with post-processing **off**, dictate the same → "deploy to Vercel" (was "deploy to Purcell").

Same three steps on each platform.

### Tests

No smoke-test check added on Windows: the change is a branch inside an async orchestrator method that the smoke harness can't construct without a full service graph, and the alternative was refactoring the pipeline purely to make it reachable. `ApplyHardenedReplacement` itself is already covered there.

The Windows side compiles clean and the existing smoke suite passes on a local .NET 10 build. The macOS side builds clean (`xcodebuild -scheme hyperwhisper -configuration Debug`, arm64). Neither change is covered by an automated end-to-end check — both were verified by reading the branch structure, so the dictation walkthrough above is worth running before merge.

### DIFF
--- a/app/macos/hyperwhisper/Managers/Transcription/Pipeline/TranscriptionPipeline+Transcription.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/Pipeline/TranscriptionPipeline+Transcription.swift
@@ -263,7 +263,14 @@ extension TranscriptionPipeline {
                 // Honor dictated break commands ("new line" / "new paragraph")
                 // in the batch path too — they were streaming-only, so with AI
                 // post-processing off they were silently dropped (issue #1).
-                finalText = TranscriptionTextProcessing.processVoiceCommands(withoutFillers)
+                let withCommands = TranscriptionTextProcessing.processVoiceCommands(withoutFillers)
+                // Vocabulary replacements are the user's own spelling corrections and
+                // are independent of any AI step, so they must apply here too (issue
+                // #80). Applied last, mirroring the two post-processing branches
+                // above; deliberately not hoisted out of the if/else, because those
+                // branches would then run replacements twice and pairs aren't
+                // guaranteed idempotent when one entry's output is another's input.
+                finalText = vocabularyProcessor.applyVocabularyReplacements(withCommands, mode: mode)
             }
 
             markStage("cache_result")
--- a/app/windows/HyperWhisper/Services/Transcription/TranscriptionOrchestrator.cs
+++ b/app/windows/HyperWhisper/Services/Transcription/TranscriptionOrchestrator.cs
@@ -270,6 +270,11 @@ public class TranscriptionOrchestrator : IDisposable
             // batch path too — they were streaming-only, so with AI post-processing
             // off they were silently dropped (issue #1).
             finalText = TranscriptionTextProcessing.ProcessVoiceCommands(finalText);
+            // Vocabulary replacements are the user's own spelling corrections, not
+            // part of AI post-processing — turning post-processing off must not
+            // discard them. Applied last, mirroring the post-processing branch
+            // above where it is the final transformation.
+            finalText = _vocabularyProcessor.ApplyReplacements(finalText);
         }
 
         return new TranscriptionResult(


ray-amjad:
Took the macOS side here rather than leaving it for a follow-up — pushed as a second commit (5bd6ea6) so it stays reviewable on its own.

Confirmed your read of `TranscriptionPipeline+Transcription.swift:253`: the `else` branch ended at `processVoiceCommands` and `applyVocabularyReplacements` had exactly two call sites in the whole app, both in the post-processing branches. So with post-processing off, nothing applied replacements on the batch path. Fixed the same way as the Windows side — last in the branch, not hoisted out of the `if`/`else`.

Builds clean on arm64 Debug. I've rewritten the PR description to cover both platforms and retitled it, and dropped the "### macOS" note since it's no longer pending. Also swapped the "Fixes #80 (Windows side)" line for plain "Fixes #80", so the issue closes on merge.

Still unverified end-to-end on either platform — the dictation walkthrough in the description is worth a run before merge.

While I was in there I checked your other two PRs against macOS for the same reason. Neither carries over: #79 depends on `ToEntity()` being a write-back over an aliased entity, and macOS History renders read-only `HistoryItemSnapshot` value types with no path back to the `Transcript`; #78 is the WPF `ListBox` internal `ScrollViewer` swallowing the wheel event, and macOS is a plain `ScrollView` + `LazyVStack`.

One thing worth noting from that pass, not for this PR: `PersistenceController.markTranscriptFailedInBackground` writes `failed` and overwrites `text` with no terminal-status guard — the same hardening your second commit on #79 adds on Windows. Nothing can currently reach it on a completed row (it's only called from genuine `catch` paths, and the launch repair is predicate-scoped to `processing`), so it's defence-in-depth rather than a live bug.

ray-amjad:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #99, opened 2026-07-31
fix(windows, macos): cap Groq post-processing output so long dictations survive

### LINKED ISSUE #98: Windows + macOS: long dictations through Groq gpt-oss silently lose post-processing — no output cap sent, Groq truncates at 2,048 tokens
# Windows + macOS: long dictations through Groq gpt-oss silently lose post-processing — no output cap sent, Groq truncates at 2,048 tokens

**Version:** 1.8.2 (x64); macOS affected by inspection (same request shape)

## Symptom

Post-processing a ~3-minute dictation (2,100 chars) through **Groq → GPT OSS 20B** fails and falls back to the raw transcript, while short dictations through the same mode succeed:

```
21:43:39.958  PostProcessingService: Processing with Groq/openai/gpt-oss-20b
21:43:42.514  PostProcessingService: Response rejected (OutputLimit); keeping original transcription
```

The completion policy is doing its job — the rejected response really is truncated — but from the user's side a long dictation just never gets cleaned up.

## Cause

Groq returns HTTP 200 with `finish_reason: "length"`. Neither app sends `max_tokens` on OpenAI-compatible requests — safe on OpenAI, where omitting it means "model maximum", but Groq's undocumented server-side default is **2,048 completion tokens**.

gpt-oss is a reasoning model whose hidden chain-of-thought spends from that same budget. Replaying the exact failed request (same transcript, prompt, and vocabulary) reproduced it 2/2: the model spent 1,734 and 2,046 of the 2,048 tokens on reasoning, leaving a truncated or completely empty visible output. Short dictations leave the reasoning small enough to fit, which is why only long ones fail. Deterministic, not a flake.

macOS builds the Groq request the same way (`AIPostProcessor` sets no output cap for cloud providers), so it fails identically.

phil-cheesman:
Fixes #98

A ~3-minute dictation (2,100 chars) through **Groq → GPT OSS 20B** came back `finish_reason: "length"`, so the completion policy rejected it (`OutputLimit`) and fell back to the raw transcript. Short dictations through the same mode kept working:

```
21:43:39.958  PostProcessingService: Processing with Groq/openai/gpt-oss-20b
21:43:42.514  PostProcessingService: Response rejected (OutputLimit); keeping original transcription
```

Neither app sends `max_tokens` on OpenAI-compatible requests — safe on OpenAI, where omitting it means "model maximum", but Groq's undocumented default is **2,048 completion tokens**, and gpt-oss spends its hidden reasoning tokens from that same budget. Replaying the exact failed request (same transcript, prompt, and vocabulary) reproduced it 2/2:

| Replay | completion_tokens | of which reasoning | visible output |
|---|---|---|---|
| no cap #1 | 2048 (exactly) | 1,734 | 1,483 chars, truncated |
| no cap #2 | 2048 (exactly) | 2,046 | empty |
| `max_completion_tokens: 4096` #1 | 3,445 | 3,021 | complete, `finish_reason=stop` |
| `max_completion_tokens: 4096` #2 | 1,736 | 1,332 | complete, `finish_reason=stop` |

### Change

Send `max_completion_tokens: 4096` on Groq requests only; every other provider keeps today's request shape. One commit per platform, reviewable separately.

Two values that look more obvious were tried and rejected against the live API:

- **8,192** (parity with the Anthropic path): Groq's free-tier TPM admission check (8,000 for gpt-oss-20b) counts **prompt + requested cap**, not actual usage — with 8,192 every request 413s before generation even starts. 4,096 accommodated every observed reasoning run with the cleaned output on top, and keeps prompt + cap admissible for dictations up to roughly seven minutes.
- **`reasoning_effort: "low"`**: collapsed reasoning to ~33 tokens and halved latency, but cleanup quality dropped to near-verbatim — fillers kept, no punctuation restored. Not worth it.

## Windows

`BuildOpenAIRequestJson` takes the `OpenAICompatibleProvider` and adds the cap for Groq. Custom OpenAI-compatible endpoints keep the uncapped shape via a provider-less overload — the server behind a user-supplied URL is unknown.

## macOS

Same cap in `AIPostProcessor`'s request assembly. All cloud providers route through the non-streaming `performAIPostProcessing` path (streaming is local-LLM-only), so the single site covers every Groq model.

### Verifying

1. Mode with post-processing set to **Groq → GPT OSS 20B**.
2. Dictate ~3 minutes of continuous speech → cleaned text is pasted (was: "Post-processing failed, using original text" toast and the raw transcript).
3. A short dictation still cleans up as before.

Done on Windows with a real 3m15s dictation through a live Groq key — cleaned text pasted where the same length previously failed.

### Tests

New Windows smoke test asserts the Groq request carries `max_completion_tokens` (and no deprecated `max_tokens`); the existing "OpenAI omits an output-token cap" check still passes with the provider argument. Full smoke suite green on a local .NET 10 build.

The macOS change structurally mirrors the Windows one but was written from a Windows machine — it's 14 additive lines against the verified `.groq` enum case, but worth an `xcodebuild` and the dictation walkthrough above before merge.

Residual limitation, deliberate: dictations much past ~5 minutes can still occasionally truncate at 4,096 (gpt-oss reasoning length is highly variable — 1,332 to 3,021 tokens across four runs on the same prompt), and free-tier admission blocks any larger cap. The completion policy's fallback to the raw transcript covers that case.

### DIFF
--- a/app/macos/hyperwhisper/Managers/Transcription/PostProcessing/AIPostProcessor.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/PostProcessing/AIPostProcessor.swift
@@ -82,6 +82,11 @@ class AIPostProcessor: ObservableObject {
     /// Max output tokens requested from any LLM (local or cloud) during post-processing.
     private let maxOutputTokens = 8_192
 
+    /// Output-token cap for Groq requests — see the comment at the Groq branch
+    /// in performAIPostProcessing's request assembly for why this is lower
+    /// than maxOutputTokens.
+    private let groqMaxCompletionTokens = 4_096
+
     /// Resolver for on-device local models
     weak var localModelManager: LocalModelManager?
 
@@ -364,6 +369,15 @@ class AIPostProcessor: ObservableObject {
         if provider == .localLLM {
             localLLMSamplingParameters.forEach { requestBody[$0.key] = $0.value }
             requestBody["max_tokens"] = maxOutputTokens
+        }
+
+        // Groq defaults to 2,048 completion tokens when the request omits a cap,
+        // and gpt-oss reasoning tokens spend from that same budget, so long
+        // dictations truncate (finish_reason=length). Kept at 4,096 — not
+        // maxOutputTokens (8,192) — because Groq's free-tier TPM admission check
+        // (8,000) counts prompt + requested cap, not actual usage.
+        if provider == .groq {
+            requestBody["max_completion_tokens"] = groqMaxCompletionTokens
         }
 
         request.httpBody = try JSONSerialization.data(withJSONObject: requestBody)
--- a/app/macos/hyperwhisper/Managers/Transcription/PostProcessing/AIPostProcessor.swift
+++ b/app/macos/hyperwhisper/Managers/Transcription/PostProcessing/AIPostProcessor.swift
@@ -371,11 +371,12 @@ class AIPostProcessor: ObservableObject {
             requestBody["max_tokens"] = maxOutputTokens
         }
 
-        // Groq defaults to 2,048 completion tokens when the request omits a cap,
-        // and gpt-oss reasoning tokens spend from that same budget, so long
-        // dictations truncate (finish_reason=length). Kept at 4,096 — not
-        // maxOutputTokens (8,192) — because Groq's free-tier TPM admission check
-        // (8,000) counts prompt + requested cap, not actual usage.
+        // Groq applies a low default completion cap when the request omits one
+        // (its reasoning docs cite 1,024), and gpt-oss reasoning tokens spend
+        // from that same budget, so long dictations truncate
+        // (finish_reason=length). Kept at 4,096 — not maxOutputTokens (8,192) —
+        // because Groq's free-tier TPM ceiling for openai/gpt-oss-120b is 8,000
+        // and the admission check counts prompt + requested cap, not actual usage.
         if provider == .groq {
             requestBody["max_completion_tokens"] = groqMaxCompletionTokens
         }
--- a/app/windows/HyperWhisper.SmokeTests/Program.cs
+++ b/app/windows/HyperWhisper.SmokeTests/Program.cs
@@ -105,13 +105,26 @@ internal static class Program
             Run("OpenAI post-processing omits an output-token cap", () =>
             {
                 var requestJson = PostProcessingService.BuildOpenAIRequestJson(
-                    "gpt-4.1-nano", "system", "user");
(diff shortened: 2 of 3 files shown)


ray-amjad:
Reviewed as part of a sibling sweep over every chat-completion request builder across macOS, Windows and the cloud backend. **The diagnosis holds and the choice of 4,096 is well-founded** — but the family has a third member this PR doesn't touch, and one number in the comment isn't right.

**Verified:**
- Groq's free-tier ceiling for `openai/gpt-oss-120b` is exactly 30 RPM / **8K TPM** ([rate limits](https://console.groq.com/docs/rate-limits)), so the 8,000 figure is correct — and it's per-model, not account-wide.
- Groq's admission check counts *requested* tokens, not actual usage: its 413 reads `Request too large … on tokens per minute (TPM): Limit 6000, Requested 12328`. So capping at 4,096 rather than 8,192 is the right call, and the reasoning in the comment is sound.
- Splitting `BuildOpenAIRequestJson` is necessary, not cosmetic: on `main` the single-arg overload is shared by Groq/Grok/Gemini/Cerebras/Mistral, so the existing "OpenAI post-processing omits an output-token cap" smoke test was silently asserting the no-cap policy for Groq too.

**One correction, which I've pushed to this branch (7e065ca):** the comment says Groq defaults to 2,048 completion tokens. Groq doesn't document that figure anywhere — its [reasoning guide](https://console.groq.com/docs/reasoning) cites **1,024** for the GPT-OSS models and the API reference states no default at all. I rewrote both comments (macOS + Windows) to describe the behaviour and cite what Groq actually documents, and named the model the 8,000 TPM ceiling belongs to. No behaviour change; the 4,096 value stands.

**The gap:** the cloud backend is the third member of this family and is still uncapped. `hyperwhisper-cloud/src/providers/groq-llm.ts` sends the same `openai/gpt-oss-120b` with no cap, and `providers/llm-request-policy.test.ts` has a passing test asserting that omission is correct — so if this merges as-is, the repo answers the same question two opposite ways, with a test pinning the wrong one.

That one also costs money: on `finish_reason=length` the completion policy correctly falls back to the raw transcript (`routes/post-process.ts:243-250`), but `deductCredits` runs unconditionally at `:280` — the user is billed for post-processing that returned their transcript unchanged.

Opened #124 for the cloud side so this PR stays client-scoped. Notably **Cerebras does not need the same treatment** despite running the same model — it documents no low default ([max output 32k free / 40k paid](https://inference-docs.cerebras.ai/models/openai-oss)), so capping it would lower a ceiling that isn't causing truncation.

Otherwise this looks good to merge.

PR #111, opened 2026-08-02
feat(macos): rebuild first-run onboarding on the Focused Task design

ray-amjad:
Follow-up to #16. That PR shipped the 8-step first-run flow; review of it produced the design handoff this implements, plus four correctness bugs that are fixed here.

## Design

Rebuilds the flow on the approved **Focused Task** direction, chosen from five explorations that all stayed inside the existing HyperWhisper product language (this is a redesign, not a rebrand):

- 760x580 shell, 8-segment progress hairline flush to the top edge
- accent glyph tile, one question, one supporting line, one card cluster
- footer holds Back, a reassurance line, and exactly one primary action
- "System Default" is the first device row; device names stay on one line
- Sound Settings is a bordered secondary action, not a link

All twelve states are implemented, including the three-way source fork (Cloud / on-device / bring-your-own-key) at the configure and setup steps.

## Structure

Onboarding logic moves out of the view into `OnboardingFlowModel`, which owns the step machine, staging and gating behind six narrow `@MainActor` protocol seams. The flow is now unit testable without permissions, Keychain, network or Core Data; `OnboardingLiveDependencies` adapts the real managers to those seams.

## Bugs fixed

1. **Set Up Later left production state mutated.** Reaching the microphone step applied the staged source config, and deferring afterward kept it. Every write is now staged and captured before it lands, and deferral restores the default Mode, any provider key that "Test API key" overwrote, and both halves of the microphone selection (the persisted preference and the open device, which diverge when the remembered mic is unplugged).
2. **Parakeet download errors never appeared.** The setup UI read only the Whisper manager's error. Both engines now feed one published error, attributed to the selected model's engine so a stale failure from the other engine cannot be misrendered.
3. **Cloud activation could write after the flow was dismissed.** The activation task is now owned and cancelled, and the commit boundary closes before cancellation so an in-flight continuation past its cancellation check still cannot write.
4. **Onboarding had no real test coverage.** 49 unit tests now cover gating, permissions, each source branch, provider validation, download failures, microphone lifecycle, rollback and completion.

## Localization

A key missing from a locale used to render as its raw identifier, because CFBundle does no per-key fallback. `String.localized` now resolves through Base, so an untranslated key degrades to readable English instead of `onboarding.welcome.task1.title`. That fix applies to every key in the app, not just these.

All 39 locales are translated and level with Base at 1006 keys; the 81 dead keys from the previous flow are removed.

## Verification

`app/macos/scripts/verify-onboarding.sh` is a reusable gate covering design conformance, localization, the four bug regressions, and repo hygiene (no prototype references, no hardcoded signing identity, no client-side entitlement bypass). It reads `DEVELOPMENT_TEAM` from the environment.

```
PASS: build green, hyperwhisperTests green, 22 checks passed, 0 warning(s).
hyperwhisperTests: 153 passed, 0 failed, 0 skipped
```

Baseline was 104 tests, so +49.

## Not covered

- **XCUITests are not run and nothing here claims UI coverage.** The runner is rejected by macOS container consent locally, before any test code executes. The gate prints an explicit note rather than reporting a meaningless pass.
- **The flow has not been visually walked end to end in the running app.** The window size is verified statically from `OnboardingStyle`, not by measuring a live window. A Debug build shares its bundle id with an installed HyperWhisper, so forcing first-run would mutate real Core Data and UserDefaults.
- Three locales (es, ja, zh-Hans) define `modes.postProcessing.privacyNotice` twice. Pre-existing on main, unrelated to this change, worth a separate cleanup.

https://claude.ai/code/session_01CG3vFXdTzju5sE4WKcofYG


ray-amjad:
## Code review — high-effort automated pass

Multi-agent review of the rebuilt onboarding flow (commits `5350054`, `54ba51f`, `6577209`). 8 correctness defects and 2 cleanup items survived adversarial verification. Ordered most-severe first.

---

### 1. Stale in-flight key test still writes the old key to the Keychain — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift:559`

`testProviderKey()` persists the probed API key to the Keychain **before** the staleness guard (`selectedProvider == provider && apiKeyInput == key`, lines 561–565), so a superseded in-flight test still writes the abandoned key; `complete()` then clears `providerKeyRestorePoints`, making the silent write permanent, and `canContinue`'s `hasKey(for:)` fallback opens the Continue gate on a key the user replaced without ever validating the replacement.

**Failure scenario:** User types OpenAI key A, clicks "Test API key", and while the probe is in flight switches the provider chip to Deepgram (which clears the field) — or edits the field to key B. The probe returns healthy and key A is persisted to the login Keychain anyway; the guard only suppresses the UI result. A key the user believed was discarded is permanently stored, and for the same-provider edit case the Continue gate opens via `hasKey` even though the key now in the field was never tested — transcription silently uses the old key A.

### 2. Parakeet download progress UI never updates — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift:95`

Parakeet download progress/state lives on `ParakeetModelManager.downloads` (a separate `DownloadController` ObservableObject) that nothing in the onboarding view tree observes.

**Failure scenario:** User picks the recommended Parakeet V2 (474 MB) and presses Download on the setup step. `DownloadController.report()` only publishes on the controller; `ParakeetModelManager.objectWillChange` fires once at start and again only at completion. The setup screen renders "Downloading… 1%" and stays frozen for the entire multi-minute download — percentage, progress bar, and Continue gate never move until it silently finishes. Whisper models are unaffected because `downloadProgress` is `@Published` directly on `WhisperModelManager` (and `ModelLibraryManager` explicitly merges `parakeetManager.downloads.$progress` for exactly this reason — confirming the asymmetry).

### 3. Credits shown after a key test belong to the anonymous device identity — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift:327`

The configure-step cloud card refreshes credits after a key probe, but `refreshCredits()` resolves the identifier via `licenseManager.getTranscriptionIdentifier()`, which returns the anonymous device id until the license is *activated* (`LicenseManager.swift:321-327`) — so the "credits available" figure shown next to a freshly tested key belongs to the unlicensed device identity, not the account the key was just verified against.

**Failure scenario:** A new user pastes a valid access key with a large prepaid balance and clicks "Test access key". The probe passes, `.task(id: flow.licenseTestPassed)` calls `refreshCredits()`, and the backend is queried with the device identifier. The card renders "0 credits available" for an account that actually has credits, making the user think their key/credits are gone before they even reach activation.

### 4. BYOK gate marks never-probed Keychain keys "validated" — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift:764`

The BYOK Continue gate was widened from `keyValidated` (old code made an inline health probe mandatory every session) to `keyValidated || providerKeys.hasKey(for: selectedProvider)`, so any pre-existing Keychain key opens both the configure and setup gates without ever being probed this session; the setup screen then renders "Key validated with \<Provider\> ✓" for a key that was never validated.

**Failure scenario:** A returning user who long ago saved an OpenAI key in Settings (since revoked/expired) re-runs onboarding and picks "Your API Key": Continue is enabled immediately, the setup step shows the key as validated and saved, and onboarding completes with the dead key as the default Mode's transcription source. The user's very first real dictation fails with a provider auth error right after a flow that just told them everything was verified.

### 5. Microphone device list is a stale snapshot; clicking a gone device fakes selection — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift:706`

`selectDevice(id:)` sets the published `selectedDeviceID` unconditionally before `LiveOnboardingAudioGateway.selectDevice` validates the id against `availableDevices` (which silently returns for an absent device), and `deviceOptions` is a one-shot snapshot taken only in `beginMicrophoneStep` — so the microphone step's device list neither shows devices plugged in while on the step nor drops unplugged ones, and clicking a stale row marks it selected in the UI while the actual input device and stored preference are unchanged.

**Failure scenario:** On the "Say something. Watch the bars." step the user unplugs their USB mic and plugs in a different headset: the new headset never appears, the old USB row is still listed, and clicking it renders the radio as selected while the gateway's guard silently does nothing — the level meter keeps metering another device and the user finishes onboarding believing an input was selected that was never applied.

### 6. Setup copy says "keep going" but the Continue gate blocks it — CONFIRMED
`app/macos/hyperwhisper/Localizations/Base.lproj/Localizable.strings:349`

`onboarding.setup.onDevice.note` ("Keep going. The download finishes in the background while you set up the microphone.") contradicts the mandatory gate: `canContinue` for `.setup` requires `catalog.isInstalled(model)` for onDevice (`OnboardingFlowModel.swift:766-780`), so Continue stays disabled until the download completes and "keep going" is impossible.

**Failure scenario:** User picks Whisper Large v3 Turbo (809 MB) and starts the download. The screen tells them to keep going while it finishes in the background, but Continue remains disabled for the entire multi-minute download; the user re-reads the copy, clicks the disabled button, and concludes the app is stuck. (Also affects the 38 translated locales of this key.)

### 7. Navigating Back to the configure step wipes a just-passed cloud key test — CONFIRMED
`app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift:230`

`OnboardingConfigureView`'s `onAppear` calls `flow.resetConfigureTestResults()`, which clears `keyValidated`; for the HyperWhisper Cloud branch `canContinue` is `license.isActive || keyValidated` with no stored-state fallback (unlike the BYOK branch), so navigating Back from the setup step before activating shuts the Continue gate on a key that was just successfully validated.

**Failure scenario:** User pastes their access key, "Test access key" passes, they Continue to the setup step, then press Back to re-read the configure screen. `onAppear` wipes `keyValidated` and Continue is now disabled even though the validated key is still sitting in the field — the user must re-run the exact same test to proceed.

### 8. Rollback can leave the app selection pointing at a deleted Mode — PLAUSIBLE
`app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift:446`

`restore()` only re-points the app's mode selection when `point.previousSelection` is non-nil; in the `modeExisted == false` branch it deletes the Mode that `apply()` created and had made the active selection via `appState.selectMode(updated, persist: true)` — so with a nil `previousSelection`, `AppState.selectedModeId`/`selectedModeSnapshot` are left referencing a deleted Core Data Mode.

**Failure scenario:** On an install where `findDefaultMode()` returned nil and no mode was selected, the user reaches the try-it step (apply creates + selects the synthetic Default mode) and then clicks "Set Up Later". Rollback deletes the created mode but leaves appState pointing at it; the next recording resolves the persisted `selectedModeId` against a deleted row.

---

### Cleanup

**9. Shell-regex regression guards duplicate the unit tests at the wrong layer — CONFIRMED**
`app/macos/scripts/verify-onboarding.sh:577` — Section 4's ~250 lines of awk/shell call-graph guards re-verify what `OnboardingFlowModelTests` already pins with real assertions, keyed to identifier spelling via fragile regexes (`skip|later|defer`, `advance|forward`, `complete|finish`). Renaming `deferSetup()` outside the regex silently disables the bug-1 guard while the script still prints PASS; an innocent new `advanceSelection()` that touches `selectMode(` hard-fails the gate.

**10. Duplicate shortcut parser diverges from HomeView's — CONFIRMED**
`app/macos/hyperwhisper/Views/Onboarding/OnboardingFocusComponents.swift:688` — `OnboardingShortcutKeys.split` re-implements the parsing that already exists as `KeyboardShortcutBadge.parseKeys` (`Views/HomeView.swift:628`). HomeView's version uppercases the primary key and handles textual "Escape"/"Space" forms while the onboarding one does not, so the same shortcut can render differently on the two screens, and any future glyph fix must be made twice.

---

https://claude.ai/code/session_01Ta4XbeLpU5m6aEXJajwhGz

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+                // that the key was verified).
+                let key = licenseKeyInput.trimmingCharacters(in: .whitespacesAndNewlines)
+                return license.isActive || keyValidated
+                    || (!key.isEmpty && key == lastValidatedLicenseKey)
+            case .yourProvider:
+                // `keyValidated` is cleared every time this step appears, so the
+                // per-session validation record keeps the gate open across Back
+                // navigation. A key that merely sits in the Keychain but was
+                // never probed this session does not count.
+                return keyValidated || validatedProviders.contains(selectedProvider)
```
ray-amjad, line 823: **HIGH · correctness · CONFIRMED** — BYOK gate passes an untested key while the Keychain keeps the old one

The BYOK Continue gate falls back to `validatedProviders.contains(selectedProvider)`, which `invalidateProviderValidation()` never clears, so editing the API key after a successful test still passes the gate while the Keychain (written only inside `testProviderKey`) still holds the *previous* key — and `keychainCard`'s `maskedKey` renders the *new* `apiKeyInput` next to a "Saved" pill.

**Failure scenario:** the user pastes OpenAI key A and presses "Test API key" — it passes, `providerKeys.persist(A)` writes A to the Keychain and `validatedProviders` gains `.openai`. They notice it is the wrong account, paste key B over it, and press Continue without re-testing. `keyValidated` is false but `validatedProviders.contains(.openai)` is true, so Continue is enabled and `isSelectedSourceUsable` is also true. The Setup step renders `maskedKey` from `apiKeyInput` (`sk-proj-B…wxyz`) beside "Key written to Keychain / Saved", and onboarding completes. Every subsequent transcription is billed to key A; key B was never stored anywhere, and the user has explicit on-screen confirmation that it was.

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+    /// which is what makes Parakeet failures visible.
+    ///
+    /// Everything here is produced INSIDE this flow. `license.lastError` and
+    /// `providerKeys.validationError` are app-global, long lived, and unobserved,
+    /// so falling back to them used to render an unrelated failure from an earlier
+    /// session before the user had done anything on the step.
+    private func refreshSetupError() {
+        switch selectedSource {
+        case .onDevice:
+            setupErrorMessage = selectedModel.flatMap { downloadErrors.message(for: $0.kind) }
```
ray-amjad, line 708: **HIGH · correctness · CONFIRMED** — download errors are keyed by engine, not by model id

Download failures are attributed by engine (`OnboardingDownloadErrors.message(for: kind)`), not by model id, so one Parakeet/Whisper model's stale failure is rendered as an error for a different model of the same engine.

**Failure scenario:** on the Set up step the user's Parakeet V2 download fails; `ParakeetModelManager.errorMessage` is set and stays set until the *next* Parakeet download starts (`ParakeetModelManager.swift:261` is the only clear). The user goes Back, picks Parakeet V3 in Configure, and returns to Set up. `refreshSetupError()` maps `selectedModel.kind == .parakeet` straight to the same string, so the V3 screen shows the V2 failure text under an untouched Download button before the user has pressed anything. The same happens between Whisper Base and Whisper Large v3 Turbo. Because the subscription in `init` fires immediately with whatever the app-global manager already holds, an error from a download attempted before the sheet opened is also presented as if it belonged to the current selection.

Same root cause also at `OnboardingSourceViews.swift:689`, `OnboardingLiveDependencies.swift:113`, `OnboardingFlowModel.swift:705`.

app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift
```diff
@@ -0,0 +1,519 @@
...
+        appState.selectMode(updated, persist: true)
+    }
+
+    func restore(_ point: OnboardingRestorePoint) {
+        guard let point = point as? LiveOnboardingRestorePoint else { return }
+
+        if !point.modeExisted {
+            // Nothing was there before, so remove what `apply` created rather than
+            // leaving a synthetic default behind.
+            if let created = persistence.fetchMode(withId: point.modeID.uuidString) {
```
ray-amjad, line 434: **HIGH · correctness · PLAUSIBLE** — Set Up Later can delete a pre-existing user Mode

`restore` deletes the Mode carrying the well-known default UUID whenever no Mode was flagged `isDefault` at capture time, but `apply` reaches that same row **by id** and may have overwritten a pre-existing user Mode rather than creating one.

**Failure scenario:** `captureRestorePoint()` finds the row via `findDefaultMode()` (predicate `isDefault == YES`) and records `modeExisted: false` when nothing is flagged default — reachable after a Local API `PATCH /modes` with `isDefault:false` (`ModesEndpoint.swift:184`) or a backup restore that clears the flag (`PersistenceController.swift:2457`) without promoting a replacement. `apply()` then calls `createOrUpdateMode(id: Self.defaultModeID, …)`, which looks the row up by id, finds the still-present seeded Mode …0001, and overwrites its name/preset/language/instructions. Pressing "Set Up Later" takes the `!point.modeExisted` branch, `fetchMode(withId:)` finds that same row, and `deleteMode` destroys it. The user's Mode and all its settings are permanently gone, with no undo — the opposite of the rollback this branch claims to perform.

app/macos/hyperwhisper/Views/Onboarding/OnboardingSourceViews.swift
```diff
@@ -86,983 +91,785 @@ struct OnboardingModelSelection: Identifiable, Equatable {
...
-                if isDownloaded {
-                    Label("onboarding.model.downloaded".localized, systemImage: "checkmark.circle.fill")
-                        .labelStyle(.iconOnly)
-                        .foregroundColor(.green)
-                } else {
-                    Text(model.size)
-                        .font(.system(size: 11))
-                        .foregroundColor(.secondary)
-                        .frame(width: 56, alignment: .trailing)
+                if let credits = hyperWhisperCloudManager.credits {
```
ray-amjad, line 301: **MEDIUM · correctness · CONFIRMED** — the Cloud "get credits" link is hidden until a key already works

The "Don't have a key? Get credits at hyperwhisper.com" link on the Cloud configure step is now rendered only after a key test has already passed, so the users it was written for can never see it.

**Failure scenario:** the old `OnboardingConfigureView.cloudConfigure` rendered `Link("onboarding.configure.cloud.getCredits")` unconditionally (old file line 497). In the rewrite the equivalent button lives inside `else if flow.licenseTestPassed == true { … if let credits = … }`. A first-run user picks HyperWhisper Cloud (the first, recommended card), lands on Configure with no key, and sees only a key field and a "Test access key" button. `canContinue` requires `license.isActive || keyValidated || key == lastValidatedLicenseKey`, so Continue is disabled and there is no longer any in-app path to the page where a key is purchased. Their only exit is "Set Up Later", which abandons the whole Cloud setup — the exact drop-off the link existed to prevent.

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+
+        audio.isRecordingPublisher
+            .sink { [weak self] recording in self?.isRecording = recording }
+            .store(in: &cancellables)
+
+        audio.transcriptPublisher
+            .sink { [weak self] text in self?.transcript = text }
+            .store(in: &cancellables)
+
+        refreshPermissions()
```
ray-amjad, line 430: **MEDIUM · correctness · CONFIRMED** — `init` publishes permission state inside the SwiftUI view-update pass

`OnboardingFlowModel.init` ends with `refreshPermissions()`, which calls `audio.refreshMicrophonePermission()` → `AudioRecordingManager.checkMicrophonePermission()` → `MicrophonePermissionManager.hasMicrophonePermission`, mirrored into `AudioRecordingManager`'s own `@Published` via `assign(to:)` (`AudioRecordingManager.swift:667`). Because the model is built by `StateObject(wrappedValue:)` in `OnboardingFlowContainer.init`, that publish happens inside SwiftUI's view-update pass. The old code did this check in `.onAppear` (`setupInitialState`), and the container already re-does it in `.onAppear`.

**Failure scenario:** the user launches the app for the first time and `appState.showOnboarding` flips true. SwiftUI allocates the container's `@StateObject` during the update pass, `AudioRecordingManager.hasMicrophonePermission`/`errorMessage` publish mid-update, and every view observing that environment object (MainAppView, HomeView, the menu-bar extra) is invalidated while it is being rendered — "Publishing changes from within view updates is not allowed, this will cause undefined behavior" — with the sheet's first frame rendered from partially-invalidated state, or a transaction assertion in release-mode SwiftUI.

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+        deviceOptions = [.systemDefault(name: systemDefaultDeviceName)] + devices
+        selectedDeviceID = audio.selectedDeviceID ?? ""
+    }
+
+    func selectDevice(id: String) {
+        // A device can vanish between the menu being built and the pick landing.
+        // Rejecting it here keeps a disconnected microphone out of the UI
+        // selection and, more importantly, stops it from flipping the pending
+        // write flag for a change that was never applied (bug 6).
+        guard id.isEmpty || deviceOptions.contains(where: { $0.id == id }) else { return }
```
ray-amjad, line 747: **MEDIUM · correctness · PLAUSIBLE** — a visually no-op device tap erases the saved microphone preference

`selectDevice(id:)` has no guard against re-selecting the row that is already highlighted, so a visually no-op tap performs a real, non-undone write to `settingsManager.selectedMicrophoneId`.

**Failure scenario:** `applyDeviceList` sets `selectedDeviceID = audio.selectedDeviceID ?? ""`, which reads `audioManager.selectedDevice?.id` — nil when the user's remembered microphone is not currently connected. The "System Default" row therefore renders as already selected while `settingsManager.selectedMicrophoneId` still names, say, their AirPods. Tapping that already-highlighted row falls through the guard (`id.isEmpty` is true), captures the restore point, and calls `audio.selectDevice(id: nil)` → `settingsManager.selectedMicrophoneId = ""` (`OnboardingLiveDependencies.swift:252`). Finishing with "Done" runs `complete()`, which discards `previousDeviceID`/`didCaptureDevice` without restoring, so the saved microphone preference is silently erased and the app falls back to the system default from then on.

app/macos/hyperwhisper/Managers/HyperWhisperCloudManager.swift
```diff
@@ -261,6 +261,24 @@ class HyperWhisperCloudManager: ObservableObject {
...
+    /// license manager currently reports.
+    ///
+    /// Onboarding needs this to show the balance of a key that has been verified
+    /// but not yet activated on this Mac, where the default identity is still the
+    /// anonymous device id.
+    ///
+    /// Like every other path through the private fetch, this publishes `credits`
+    /// and `lastFetchTime` for the override identity; both are corrected by the
+    /// next default refresh once the license is activated.
+    func refreshCredits(identifierOverride: String) async {
```
ray-amjad, line 275: **MEDIUM · correctness · PLAUSIBLE** — a merely-probed key's credits are published app-wide

`refreshCredits(identifierOverride:)` publishes an arbitrary probed key's balance into the app-wide `credits` and refreshes `lastFetchTime`, breaking the invariant that `credits` describes the identity the license manager reports.

**Failure scenario:** every previous public entry point resolved the identity from `licenseManager.getTranscriptionIdentifier()`; the only override caller (`refreshServerLicenseCache`) is used for the key that is already transcribing. The new method is called from `OnboardingConfigureView.cloudCard`'s `.task(id: flow.licenseTestPassed)` with whatever string the user typed into the access-key field — a key that has explicitly *not* been activated on this Mac. Inside the private `fetchCredits`, the override path still executes `self.credits = fetchedCredits` and `self.lastFetchTime = Date()`. So: a user already licensed under key A has onboarding re-shown, pastes a colleague's key B to see what it does, the probe passes, and `credits` app-wide is now key B's balance with a fresh 60s cache stamp. They press "Set Up Later", which by design does not roll back anything cloud-side. Until something force-refreshes, `CloudAccountSettingsSection` shows key B's remaining credits, minutes and currency as the user's own, and drives the low-credit / exhausted warning banners off that foreign number.

app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift
```diff
@@ -0,0 +1,519 @@
...
+                foreignPlatformExtensions: point.foreignPlatformExtensions
+            )
+            if restored.isDefault != point.wasDefault {
+                restored.isDefault = point.wasDefault
+                persistence.save()
+            }
+        }
+
+        // Put the active mode selection back where it was.
+        if let previous = point.previousSelection {
```
ray-amjad, line 471: **MEDIUM · correctness · PLAUSIBLE** — rollback skips Mode re-selection when the snapshot is nil

`restore()` only re-selects the active Mode when `point.previousSelection` is non-nil, but that value is `appState.selectedModeSnapshot`, which `AppState.refreshSelectedModeSnapshot` fills from a *background* fetch (`AppState.swift:920`) and nils out for an empty id — so when the snapshot has not landed yet, "Set Up Later" restores the Mode row's fields but leaves `selectedModeId` / `settingsManager.currentModeId` pointing at the Default Mode that `apply()` selected.

**Failure scenario:** a returning user with `onboardingPending` set has their custom Mode "Email" active. Onboarding opens before the async `fetchModeSnapshotInBackground` for "Email" resolves (or after it was cleared), so `captureRestorePoint()` records `previousSelection: nil` while `modeExisted` is true. The user reaches the Try It step (which runs `apply()` → `appState.selectMode(Default, persist: true)`), then presses "Set Up Later". Rollback rewrites the Default Mode's fields but skips the re-selection branch, so the app — and `settingsManager.currentModeId`, which persists across launches — is now on "Default" instead of "Email". The next dictation runs with Default's language, prompt and post-processing settings, contradicting the "puts the app back exactly as it was" contract.

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+
+    /// The mandatory gate on the setup step: is the chosen source genuinely
+    /// usable right now.
+    var isSelectedSourceUsable: Bool {
+        guard let source = selectedSource else { return false }
+        switch source {
+        case .onDevice:
+            guard let model = selectedModel else { return false }
+            return catalog.isInstalled(model)
+        case .hyperwhisperCloud:
```
ray-amjad, line 840: **MEDIUM · correctness · PLAUSIBLE** — onboarding no longer observes `LicenseManager`

The onboarding views no longer observe `LicenseManager`, so a `licenseStatus` change that does not originate from `activateCloudLicense()` never re-renders the Cloud setup card or re-evaluates the mandatory Continue gate.

**Failure scenario:** the old `OnboardingView` and `OnboardingSetupView` both held `@EnvironmentObject var licenseManager`, so every `LicenseManager` publish re-ran `canContinueFromCurrentStep()` and re-rendered `cloudSetup`. The rewrite reads license state only through `OnboardingLicenseGateway.isActive` (`manager.licenseStatus == .active`) from a plain `@MainActor` adapter, and `OnboardingFlowModel` sets up Combine sinks for the catalog, devices, recording and transcript but none for `LicenseManager`; `OnboardingSetupView` now observes only `flow` and `hyperWhisperCloudManager`. `LicenseManager.init` kicks off `loadStoredLicense()` at launch (the sheet appears 0.5s later) and `processValidationResult` writes `licenseStatus` from background revalidation. So: a returning licensed user sits on the Setup step while launch revalidation flips `licenseStatus` to `.active` — the card keeps showing the "Activate Cloud" button and Continue stays disabled at the mandatory gate until some unrelated `@Published` on the flow happens to fire. The mirror case is worse: a revalidation that downgrades to `.expired` leaves the gate reading a stale `.active` and lets the user through Setup on a license the server has already rejected.

app/macos/scripts/verify-onboarding.sh
```diff
@@ -0,0 +1,680 @@
...
+
+# --- locale coverage, information only -------------------------------------
+BASE_N=$(grep -c . "$DEFINED")
+LAGGING=""
+for d in "$LOC_DIR"/*.lproj; do
+    name=$(basename "$d" .lproj)
+    [ "$name" = "Base" ] && continue
+    f="$d/Localizable.strings"
+    [ -f "$f" ] || { LAGGING="$LAGGING $name:none"; continue; }
+    n=$(grep -cE '^[[:space:]]*"onboarding\.' "$f" 2>/dev/null || echo 0)
```
ray-amjad, line 544: **LOW · cleanup · CONFIRMED** — `grep -c` with `|| echo 0` yields two values and masks an untranslated locale

`n=$(grep -cE ... || echo 0)` emits two values when grep matches nothing, because `grep -c` already prints `0` before exiting non-zero.

**Failure scenario:** a locale whose `Localizable.strings` contains no `onboarding.*` key sets `n` to the two-line string `0\n0`; the next line's `[ "$n" -ne "$BASE_N" ]` fails with "integer expression expected" on stderr and evaluates false, so the completely-untranslated locale is silently reported as level with Base instead of being listed as lagging.

Same pattern at lines 359 and 381.

ray-amjad (commented):
Automated code review (high effort, workflow-backed: 4 finder angles + independent verify pass; 36 candidates → 31 kept, 5 refuted, 10 reported).

The onboarding rewrite is structurally solid, but state leaks across its new seams. The three most serious: the BYOK Continue gate accepts an edited-but-untested key while the Keychain still holds the previous one, on-device download errors are keyed by engine rather than model id, and the Set Up Later rollback deletes whatever Mode happens to carry the default UUID. Several other rollback/observation gaps follow the same pattern — app-global state read or written from inside the flow with no per-flow scoping.

Each comment is labelled **CONFIRMED** (verifier reproduced the mechanism end to end) or **PLAUSIBLE** (mechanism verified, trigger path narrower or partly inferred).

app/macos/hyperwhisper/Views/Onboarding/OnboardingFlowModel.swift
```diff
@@ -0,0 +1,1000 @@
...
+        providerTestHealth = nil
+        providerErrorMessage = nil
+        let task = Task { [weak self] in
+            guard let self else { return }
+            let health = await self.providerKeys.probe(provider, apiKey: key)
+            guard !Task.isCancelled, self.isLive else { return }
+            // Drop a result the user has since superseded BEFORE the persist:
+            // a stale probe must never write the Keychain or set a restore
+            // point (which would also wrongly flag a pending production write).
+            guard self.selectedProvider == provider,
```
ray-amjad, line 595: **HIGH · persistence/concurrency — Cancel an abandoned provider probe before it writes Keychain**

This staleness guard does not check `selectedSource`. Start a BYOK probe, go Back, and select On-device before the request returns: `selectedProvider` and `apiKeyInput` are unchanged, so the late healthy result still captures a restore point and persists the key. If the user then completes the on-device flow, `complete()` removes all provider restore points without restoring this abandoned-branch write, leaving an API key the final configuration never selected. Cancel the probe when leaving BYOK (and include `selectedSource == .yourProvider` in this guard), then commit or discard provider restore points according to the final selected source.

app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift
```diff
@@ -0,0 +1,519 @@
...
+        return OnboardingLicenseOutcome(
+            isValid: result.isValid,
+            errorMessage: result.isValid ? nil : (result.errorMessage ?? "app.unknown.error".localized)
+        )
+    }
+
+    /// The single explicit activation. Entitlement is verified server side; there
+    /// is no local shortcut here and there must never be one.
+    func activate(_ key: String) async -> OnboardingLicenseOutcome {
+        let result = await manager.activateLicense(key)
```
ray-amjad, line 167: **HIGH · cancellation/state — Cancellation is observed only after `LicenseManager` mutates global state**

`finish()` cancels the flow task, but this awaited call is stateful. `LicenseNetworkService.performValidation` converts `CancellationError` into an `.invalid` `LicenseValidationResult`, and `LicenseManager.activateLicense` then unconditionally calls `processValidationResult`, publishing `.invalid` and clearing customer metadata before control returns to the flow's `Task.isCancelled`/`isLive` guard. Thus Set Up Later during activation can disturb an existing active license even though the flow drops the outcome; if the response wins the race, an abandoned key can instead persist. Cancellation/current-operation validation needs to happen before the manager writes shared license state, and activation should also be invalidated when leaving the Cloud branch.

app/macos/hyperwhisper/Views/Onboarding/OnboardingLiveDependencies.swift
```diff
@@ -0,0 +1,519 @@
...
+        guard let point = point as? LiveOnboardingRestorePoint else { return }
+
+        if !point.modeExisted {
+            // Nothing was there before, so remove what `apply` created rather than
+            // leaving a synthetic default behind.
+            if let created = persistence.fetchMode(withId: point.modeID.uuidString) {
+                persistence.deleteMode(created)
+            }
+        } else {
+            let restored = persistence.createOrUpdateMode(
```
ray-amjad, line 438: **MEDIUM · rollback — Restore the captured Mode without re-normalizing nullable fields**

`createOrUpdateMode` is not an exact restoration API: it converts captured nils to current defaults (`customInstructions` to `""`, `languageModel` to the default model, `cloudProvider` to `hyperwhisper`, and nil tier/provider/model fields to their current defaults). Therefore a legacy or imported Default Mode with valid nullable fields is still changed after Set Up Later, despite the byte-for-byte rollback contract. Restore the managed object's captured properties directly (or add a dedicated exact-restore API) and cover the live committer with an in-memory Core Data test; the current `FakeCommitter` rollback tests cannot expose this adapter mismatch.

ray-amjad (commented):
Three additional verified findings from the exact-head review. Two other findings were already covered by existing inline threads, so they are not duplicated here.

Issue #77, opened 2026-07-30
Windows: local LLM post-processing is CUDA-only, so AMD/Intel GPUs fall back to CPU (~29s for a 5s clip)


phil-cheesman:
# Windows: local LLM post-processing is CUDA-only, so AMD/Intel GPUs fall back to CPU (~29 s for a 5 s clip)

**Version:** 1.8.2 (x64). GPU: AMD Radeon RX 6900 XT. CPU: Ryzen (16 threads), 64 GB RAM.

## Symptom

A fully local mode — Whisper large-v3-turbo + Gemma 4 E2B post-processing — takes ~33 seconds end to end for a 5.7-second recording. Almost all of it is post-processing:

```
21:57:08.468    Backend: GPU (AMD Radeon RX 6900 XT)
21:57:08.468    Model: ggml-large-v3-turbo.bin
21:57:09.134  FILE TRANSCRIPTION COMPLETE   Inference time: 659ms
21:57:09.145  PostProcessingService: Processing with LocalLlm/gemma-4-E2B-it-Q4_K_M.gguf
21:57:12.464  LocalLlmService: Loaded gemma-4-E2B-it-Q4_K_M.gguf (CPU fallback)
21:57:28.422  [PERF] Transcribing still pending after 20s (mode=local, lang=en)
21:57:41.150  PostProcessingService: Successfully processed (52 -> 52 chars)
21:57:41.152  TranscriptionFlow: Transcription completed in 32686ms
```

Whisper is fine: **659 ms** for 5.7 s of audio on the Radeon (and 1.9 s for a 37 s clip, 4.1 s for a 128 s clip). The **28.7 seconds** is Gemma running on CPU to rewrite 52 characters.

## Cause

`Services/LocalLlmGpuHelper.cs:32`:

```csharp
var willTryCuda = gpu.Name.Contains("NVIDIA", StringComparison.OrdinalIgnoreCase)
    && PlatformHelper.HasLocalLlmCudaRuntime;
return new RuntimePlan(
    WillTryCuda: willTryCuda,
    GpuLayerCount: willTryCuda ? 99 : 0,
    BackendSummary: willTryCuda ? "CUDA first, CPU fallback" : "CPU fallback",
    SharesGpuWithWhisper: willTryCuda);
```

Any non-NVIDIA adapter gets `GpuLayerCount: 0`, so llama.cpp offloads nothing. Consistent with the csproj, which references only:

```
[...]

PR #132, opened 2026-08-04
feat(windows): Vulkan backend for local LLM post-processing

ray-amjad:
## Summary

Fixes the "local LLM post-processing is CUDA-only" gap from #77: an RX 6900 XT machine ran Gemma post-processing on CPU (~29 s for a 5 s clip) while Whisper itself offloads fine via Vulkan. This adds `LLamaSharp.Backend.Vulkan` and makes the plan/pin machinery backend-aware.

**Addresses #77** — intentionally not auto-closing: the plan calls for a pre-release installer on the issue so phil-cheesman can validate on the RX 6900 XT first. **Supersedes #131** (the bot's badge-only PR — the badge here is built on the new backend-aware, pin-aware plan instead of the CUDA-only one).

## Design (settled in spec interview)

- **Own probe, forced path**: eligibility = ML-class discrete adapter (GpuInfoService priority 1–3) + shipped `runtimes/win-x64/native/vulkan/` files + `System32\vulkan-1.dll`. LLamaSharp's own Vulkan detection spawns the Vulkan SDK's `vulkaninfo` (absent on end-user machines), so `LocalLlmService` pins the vulkan library explicitly via `NativeLibraryConfig.All.WithLibrary(...)` on first load. The choice is process-wide and one-shot; later models that disagree re-plan against the frozen backend (usually 0-layer degrade, today's behaviour) and get their preferred backend on next launch.
- **Eligibility**: NVIDIA → CUDA first, Vulkan when CUDA is crash-pinned or cuda12 files absent; discrete AMD/Intel (RX/PRO/VEGA, Arc) → Vulkan; iGPU/APU/unknown → CPU. No VRAM arithmetic. Pure `DecideBackend` function, SmokeTest-covered.
- **Per-backend crash pins**: `GpuLoadCrashEntry.Backend` (`cuda`/`vulkan`, null = legacy → pins all backends, preserving pre-upgrade meaning). A CUDA crash no longer blocks Vulkan next launch. All documented in-flight / launch-ordinal / clean-exit invariants preserved; sentinel writes/clears are backend-scoped.
- **`ggml-cpu.dll` copy target** (build + publish) into `vulkan/` **and** `cuda12/` — verified from the actual nupkgs that both GPU payloads ship without it; the pinned Vulkan path resolves every dependency from its own directory, so this is required for the Vulkan load itself, not just the CPU degrade.
- **UI**: pin-aware "GPU · CUDA" / "GPU · Vulkan" / "CPU" chip on local-LLM Model Library rows; "Recommended" tag + display-name suffix hidden when the plan is CPU-only (`GetDefault()` unchanged).
- **Diagnostics**: Sentry gains `backend` tag and `recovery = {backend}_to_{next}`; dedupe key includes the transition.
- **Strings**: `cudaRequiresNvidia` (false once AMD offloads) replaced by vendor-neutral `gpuUnsupported` + new `vulkanFirst`, translated across all 40 resx files via localisation-syncer.
- Docs (`mintlify-help/performance-tips.mdx`, `models.mdx`) reworded to "NVIDIA (CUDA), discrete AMD/Intel (Vulkan); integrated GPUs and ARM64 use CPU".

## Verification done (macOS cross-check + CI)

- Full app + SmokeTests compile clean (0 errors) with `EnableWindowsTargeting`.
- `dotnet publish -r win-x64`: `runtimes/win-x64/native/vulkan/` contains all 6 DLLs incl. injected `ggml-cpu.dll`; `cuda12/` gains `ggml-cpu.dll`; no linux/osx assets.
- `dotnet publish -r win-arm64`: no vulkan payload anywhere (only `runtimes/win-arm64`).
- SmokeTests: plan matrix (RX 6900 XT → Vulkan, APU/UHD → CPU, Arc → Vulkan, NVIDIA CUDA/pin/fallback permutations, loader/runtime absence), GPU-name → ML-eligibility classes, per-backend pin scoping incl. legacy-null semantics. windows-ci runs these on this PR.

## Still to do (Windows hardware)

1. Run the app on a real Windows box; confirm the log line reports the right backend and the badge matches.
2. `build-release.ps1 -Architecture x64` pre-release installer (~+20 MB expected) posted on #77 for phil-cheesman: expect `Loaded … (Vulkan, AMD Radeon RX 6900 XT …)`, post-processing seconds not ~29 s, no crash loop across restarts.
3. Post-release: watch Sentry `component=local_llm` for the new backend tags.

https://claude.ai/code/session_012aysnGv8HFbhNHhygkP24K


app/windows/HyperWhisper/Services/LocalLlmService.cs
```diff
@@ -225,6 +270,63 @@ private async Task LoadModelInternalAsync(string modelPath, CancellationToken ca
...
+                }
+                catch (Exception ex)
+                {
+                    LoggingService.Warn(
+                        $"LocalLlmService: Failed to pin the Vulkan native library, keeping default backend selection: {ex.Message}");
+                    desiredBackend = LocalLlmBackend.None;
+                }
+            }
+
+            _processNativeBackend = desiredBackend;
```
ray-amjad, line 310: **High · CONFIRMED · correctness** — Frozen process backend `None` forces every later model onto CPU

`PinProcessNativeLibraryOnce` sets `_processNativeBackend = desiredBackend` unconditionally on this line, but only calls `NativeLibraryConfig.All.WithLibrary(...)` inside the `Vulkan` branch. When the first load of the session plans CPU, `None` is stored as the "frozen" backend although nothing was actually pinned.

**Failure:** NVIDIA machine, two Modes with different Gemma models. The first load plans CPU — the model carries a durable crash pin, or `GetRuntimePlanSafely` swallowed a transient `GpuInfoService`/WMI failure and returned `CpuPlan(null)`. The user switches to the second model, which has no pin: `pinnedPlan.Backend` is `Cuda`, but `processBackend` (`None`) differs, so the plan is recomputed with `restrictToBackend = None`. Both `restrictToBackend is null or Cuda` and `... or Vulkan` are false, so `GpuLayerCount` is 0. The healthy model runs post-processing on CPU — tens of seconds instead of a few — until the app restarts.

Before this PR the crash pin was per-model, so the second model still loaded on CUDA. Same root cause at `LocalLlmService.cs:205`.

app/windows/HyperWhisper/Services/LocalLlmService.cs
```diff
@@ -167,54 +177,89 @@ private async Task LoadModelInternalAsync(string modelPath, CancellationToken ca
...
-            ClearGpuLoadSentinel(modelPath);
-            LoggingService.Warn("LocalLlmService: GPU load failed, retrying with CPU backend");
+            if (backendId != null)
+            {
+                ClearGpuLoadSentinel(modelPath, backendId);
+            }
+            LoggingService.Warn(
+                $"LocalLlmService: {LocalLlmGpuHelper.BackendDisplayName(plan.Backend)} load failed, retrying with CPU backend");
             UnloadModel();
             var parameters = CreateParameters(modelPath, gpuLayerCount: 0);
```
ray-amjad, line 263: **High · PLAUSIBLE · correctness** — Pinning the Vulkan `llama.dll` leaves the CPU retry no fallback

`NativeLibraryConfig.All.WithLibrary` (line 298) disables the LLamaSharp auto-fallback chain, so this `catch when (gpuLayerCount > 0)` CPU retry resolves the same library that just failed.

**Failure:** an AMD RX machine where `vulkan-1.dll` exists — so `HasSystemVulkanLoader` passes — but `runtimes/win-x64/native/vulkan/llama.dll` cannot load, because the `ggml-cpu.dll` copy did not run (see the `PlatformHelper.cs:76` finding) or the ICD/driver is broken. `LLamaWeights.LoadFromFile` throws `DllNotFoundException`/`TypeInitializationException`. The catch block calls `LoadFromFile` again with `gpuLayerCount 0`, hits the same pinned library, and throws the identical (or cached `TypeInitialization`) exception out of `LoadModelInternalAsync`.

Local post-processing then fails on every dictation for the whole session with a hard error. The same machine got working CPU post-processing before this PR.

app/windows/HyperWhisper/Services/LocalLlmService.cs
```diff
@@ -380,7 +498,18 @@ private static long CurrentLaunchOrdinal(GpuLoadCrashStore store)
...
+    /// pins <paramref name="plannedBackend"/>. A null entry backend predates
+    /// per-backend pins (shipped #563 store / legacy sentinel migration) and
+    /// pins every backend — its crash was recorded when the app only ever
+    /// forced CPU, so matching everything preserves that pre-upgrade meaning.
+    /// </summary>
+    internal static bool CrashEntryPinsBackend(string? entryBackend, string plannedBackend)
+        => entryBackend == null
+           || string.Equals(entryBackend, plannedBackend, StringComparison.OrdinalIgnoreCase);
+
+    internal static bool GpuLoadCrashedPreviously(string modelPath, string gpuIdentity, string backend)
```
ray-amjad, line 511: **High · CONFIRMED · correctness** — Models page views burn the 10-launch crash-pin budget

`GpuLoadCrashedPreviously` calls `CurrentLaunchOrdinal`, which increments and persists `store.LaunchOrdinal` (lines 494-497). This read is now reachable from `LocalLlmGpuHelper.GetRuntimePlan`, so from the Models settings page and the local API `/models` endpoint.

**Failure:** a driver kills the process with an AccessViolation during GPU weight upload, so a crash pin is written at launch N. Over the next ten launches the user opens Settings → Models, or an MCP client calls `list_models` (`ModelsEndpoints.cs:29` → `Rebuild()` → `BuildLocalLlmRows` → `GetRuntimePlan(path)` → `GpuLoadCrashedPreviously` → `CurrentLaunchOrdinal`), and never runs local post-processing. Each of those launches bumps and saves the ordinal, so at launch N+10 the pin satisfies `now - entry.LaunchOrdinal >= CrashPinExpiryLaunches` and is deleted. The next dictation retries the GPU load and the process is hard-killed mid-dictation, losing the in-flight recording — the crash loop the pin exists to prevent.

Before this PR only launches that called `LoadModelInternalAsync` advanced the counter. This also falsifies the documented invariant on `ClearInFlightGpuLoads` ("No load was attempted this run (ordinal never assigned)"), which now assigns an ordinal from page views alone.

app/windows/HyperWhisper/Services/ModelLibraryManager.cs
```diff
@@ -193,13 +193,17 @@ private IEnumerable<LibraryModel> BuildLocalLlmRows()
...
-        var runtimeGuidance = BuildLocalLlmRuntimeGuidance(runtimePlan);
         foreach (var model in LocalLlmModelInfo.AllModels)
         {
             var installed = _localLlm.IsModelDownloaded(model);
             var unsupportedReason = PlatformHelper.SupportsLocalLlmPostProcessing ? null : LocalLlmUnsupportedReason();
             var status = OfflineStatus(installed, unsupportedReason);
+            // Per-model, pin-aware plan: a backend that previously crashed the
+            // process for this model is excluded, so the badge never claims an
+            // offload the load path would refuse.
+            var runtimePlan = LocalLlmGpuHelper.GetRuntimePlan(_localLlm.GetModelPath(model));
```
ray-amjad, line 204: **Medium · CONFIRMED · correctness** — In-flight load sentinel is misread as a prior crash

This per-model pin-aware plan reads the same crash store that `WriteGpuLoadSentinel` populates. An `InFlight=true` entry written by a load that is running right now is indistinguishable from a crash pin.

**Failure:** the user triggers post-processing — `LoadModelInternalAsync` writes an `InFlight=true` entry for the model + GPU + `"vulkan"` — and, while the multi-GB weights load, opens Settings → Models (or a Local API client hits `/models`, which also calls `Rebuild()`). `GpuLoadCrashedPreviously` matches that in-flight entry (same GPU identity, same launch ordinal, `InFlight != false`) and returns true. The row renders the "CPU" chip, "Runtime: CPU fallback", the `gpuUnsupported` sentence, and silently drops the "Recommended" tag — for a model that is in fact about to run on Vulkan.

app/windows/HyperWhisper/Services/ModelLibraryManager.cs
```diff
@@ -259,9 +274,12 @@ private static string BuildLocalLlmRuntimeGuidance(LocalLlmGpuHelper.RuntimePlan
...
         }
 
-        var guidance = plan.WillTryCuda
-            ? Loc.S("settings.models.localLlm.hardware.cudaFirst")
-            : Loc.S("settings.models.localLlm.hardware.cudaRequiresNvidia");
+        var guidance = plan.Backend switch
+        {
+            LocalLlmBackend.Cuda => Loc.S("settings.models.localLlm.hardware.cudaFirst"),
+            LocalLlmBackend.Vulkan => Loc.S("settings.models.localLlm.hardware.vulkanFirst"),
+            _ => Loc.S("settings.models.localLlm.hardware.gpuUnsupported")
```
ray-amjad, line 281: **Medium · CONFIRMED · correctness** — `gpuUnsupported` asserts a permanent hardware limit for recoverable states

The `LocalLlmBackend.None` arm maps to `settings.models.localLlm.hardware.gpuUnsupported` — "CPU fallback. This GPU does not support accelerated local post-processing" — but that arm is also reached for a crash-pinned model, a missing system Vulkan loader, and the frozen-process-backend degrade. In all three the GPU is perfectly capable.

**Failure:** a Radeon RX 7900 XTX owner whose model was pinned after one bad load, or whose `vulkan-1.dll` is missing from an incomplete driver install, opens Settings → Models and reads that their GPU is unsupported. The claim is false, so they stop troubleshooting — no driver reinstall, no waiting out the 10-launch pin expiry — and conclude the card cannot do it.

app/windows/HyperWhisper/Utilities/PlatformHelper.cs
```diff
@@ -73,6 +73,39 @@ public static class PlatformHelper
             "cuda12",
             "llama.dll"));
 
+    public static bool HasLocalLlmVulkanRuntime =>
```
ray-amjad, line 76: **Medium · PLAUSIBLE · correctness** — The Vulkan runtime probe misses the required `ggml-cpu.dll`

`HasLocalLlmVulkanRuntime` checks only `vulkan/llama.dll`, but the new pinned-directory load also needs `ggml-cpu.dll` in that same folder.

**Failure:** `CopyLlamaCpuBackendIntoGpuRuntimeDirsOnBuild` (`HyperWhisper.csproj:273`) is gated on `Exists('$(OutDir)runtimes\win-x64\native\avx2\ggml-cpu.dll')` in its outer condition and `Exists('$(OutDir)runtimes\win-x64\native\vulkan')` on each `<Copy>`. If the LLamaSharp native folders are not yet materialised in `$(OutDir)` when the target runs — first clean build, incremental build ordering, any package layout change — both copies are skipped with no diagnostic at all, unlike the Publish target, which emits a `<Warning>`.

This probe still returns true because `vulkan/llama.dll` exists, so `DecideBackend` plans Vulkan and `PinProcessNativeLibraryOnce` pins a library that resolves dependencies from its own directory and cannot find a CPU compute backend. Combined with the `LocalLlmService.cs:264` finding, the AMD RX user gets local post-processing that never works at all, instead of the CPU fallback the removed code guaranteed.

mintlify-help/performance-tips.mdx
```diff
@@ -156,7 +156,7 @@ Post-processing is an optional second step that cleans up filler words, fixes pu
     Start with **Gemma 4 E2B** — it fits comfortably in 4 GB and handles most cleanup tasks well.
   </Tab>
   <Tab title="Windows">
-    Local Gemma post-processing is available on Windows x64. GPU acceleration uses NVIDIA CUDA when a compatible GPU is present; AMD and Intel GPU systems fall back to CPU.
+    Local Gemma post-processing is available on Windows x64. GPU acceleration uses NVIDIA CUDA and discrete AMD/Intel GPUs via Vulkan; integrated GPUs and ARM64 systems run on CPU.
```
ray-amjad: **Medium · CONFIRMED · correctness** — Docs overclaim discrete AMD/Intel Vulkan support

This line says GPU acceleration covers "discrete AMD/Intel GPUs via Vulkan" and that only integrated GPUs and ARM64 run on CPU. `IsDiscreteForMl` is `PriorityScore is >= 1 and <= 3`, and the AMD branch of `GetGpuPriorityScore` returns 4 for discrete parts whose names lack `" RX"`, `"PRO "`, or `"VEGA"` — so they are excluded.

**Failure:** a Radeon HD 7970 / R9 390 / FirePro owner reads the promise and then finds "CPU" on the Models page with no way to enable offload. The PR's own smoke test asserts `!EligibleForMl("AMD Radeon HD 7970")`.

The same unqualified claim appears at `mintlify-help/models.mdx:192` ("GPU acceleration on x64 uses NVIDIA CUDA and discrete AMD/Intel GPUs via Vulkan") and `mintlify-help/models.mdx:252`, plus the table row nearby at `performance-tips.mdx:169`. All need the same qualification.

app/windows/HyperWhisper/Models/LibraryModel.cs
```diff
@@ -45,6 +45,11 @@ public sealed class LibraryModel
     public required LibraryModelSource Source { get; init; }
     public string? SizeDescription { get; init; }
     public string? Tag { get; init; }
+    /// <summary>
+    /// Runtime backend chip for local-LLM rows ("GPU · CUDA", "GPU · Vulkan",
+    /// "CPU") — pin-aware per model. Null on every other row kind.
+    /// </summary>
+    public string? RuntimeBadge { get; init; }
```
ray-amjad, line 52: **Low · CONFIRMED · correctness** — `WithDownloadState` drops the new `RuntimeBadge`

`ModelsSettingsPage.xaml.cs` `WithDownloadState` rebuilds `LibraryModel` field by field — `Id`, `Tag`, `Detail`, `DetailToolTip`, `Payload` and so on — and does not copy this new `RuntimeBadge` init property.

**Failure:** the user starts a Gemma 4 E2B download. `RebuildLibrary`/`SetRowDownloading` replaces the row's model with `WithDownloadState(...)`, `RuntimeBadge` becomes null, `RuntimeBadgeVisibility` flips to `Collapsed`, and the "GPU · Vulkan" / "CPU" chip disappears from that row for the whole download. It reappears when the library is rebuilt after completion.

app/windows/HyperWhisper/Services/LocalLlmGpuHelper.cs
```diff
@@ -1,41 +1,175 @@
...
+    internal static LocalLlmBackend DecideBackend(
+        string gpuName,
+        bool isDiscreteForMl,
+        bool hasCudaRuntime,
+        bool hasVulkanRuntime,
+        bool hasSystemVulkanLoader,
+        bool cudaPinned,
+        bool vulkanPinned)
+    {
+        if (gpuName.Contains("NVIDIA", StringComparison.OrdinalIgnoreCase)
```
ray-amjad, line 125: **Low · PLAUSIBLE · cleanup** — A second NVIDIA name rule can route NVIDIA parts to Vulkan

`DecideBackend` re-tests the GPU name for `"NVIDIA"` instead of reusing the vendor classification `GetGpuPriorityScore` already performs.

`GetGpuPriorityScore` gives score 1 to GEFORCE/RTX/GTX/QUADRO/TESLA/TITAN names even when the string `"NVIDIA"` is absent; `gpuName.Contains("NVIDIA")` does not. An adapter reported as "Quadro P2000" or "GeForce RTX 4060" without a vendor prefix is therefore classified ML-discrete, skips CUDA, and is silently routed to the slower Vulkan backend.

Two name-matching rules for one vendor question will keep drifting. Expose the vendor from the existing classification (`GpuInfoService.cs:143`) instead.

app/windows/HyperWhisper/Services/ModelLibraryManager.cs
```diff
@@ -215,7 +219,18 @@ private IEnumerable<LibraryModel> BuildLocalLlmRows()
                 SizeDescription = model.Size,
                 Speed = 3,
                 Accuracy = model.IsRecommended ? 4 : 3,
-                Tag = model.IsRecommended ? "Recommended" : null,
+                // "Recommended" implies GPU offload; on CPU-fallback machines the
+                // recommendation would be misleading (~29 s for a 5 s clip), so
+                // hide it when the plan cannot offload.
+                Tag = model.IsRecommended && offloads ? "Recommended" : null,
+                RuntimeBadge = unsupportedReason == null
```
ray-amjad, line 226: **Low · CONFIRMED · cleanup** — Backend badge text is duplicated and hardcoded in English

This `RuntimeBadge` switch re-implements `LocalLlmGpuHelper.BackendDisplayName`. Four places now map `LocalLlmBackend` to text: `BackendId`, `BackendDisplayName`, the two `BackendSummary` literals, and this switch — adding a backend means editing all four, and one will be missed.

It also renders "GPU · CUDA" / "CPU" untranslated on the Models page in all 39 non-English locales, right next to guidance text that is localized with `Loc.S` on the same row.

Suggested: `$"GPU · {LocalLlmGpuHelper.BackendDisplayName(plan.Backend)}"` with a resx-backed "CPU" label.

ray-amjad (commented):
## Code review — high effort

**10 findings**, all anchored inline below. Method: 4 finder angles over the 53 changed files → 33 candidates → an independent verifier per location → 31 verified, 2 refuted.

Three root causes dominate the new Vulkan plumbing:

1. **The process-wide backend freeze** stores `None` from the first CPU-planned load, then forces every other model in the session onto CPU — a regression from the previously per-model crash pin.
2. **Pinning the Vulkan `llama.dll`** removes the LLamaSharp fallback chain, so the GPU→CPU retry re-enters the same library. Only a build-time `Warning` and an `llama.dll`-only probe guard the required `ggml-cpu.dll`.
3. **Making `GetRuntimePlan` crash-store-aware** wires a mutating read into the Models page and the local `/models` endpoint, where it burns the 10-launch pin-expiry budget on page views and misreads the current load's in-flight sentinel as a prior crash.

The rest are user-facing (the `gpuUnsupported` string, the docs' blanket AMD/Intel claim, the badge dropped during downloads) plus two maintainability items.

**Refuted by the verify pass**, listed so they are not re-raised:

- `app/windows/HyperWhisper/Services/LocalLlmGpuHelper.cs:86` — the Cuda and Vulkan switch arms are near-duplicates. Style only, no observable effect.
- `app/windows/HyperWhisper/HyperWhisper.csproj:291` — `ggml-cpu.dll` is also copied into `cuda12/`. Intentional per the PR's own comment.

mintlify-help/models.mdx
```diff
@@ -189,7 +189,7 @@ Local **Gemma 4** models clean up and format transcript text fully offline after
     | Gemma 4 31B Dense | 18.3 GB | ~20 GB | Highest local quality, slowest |
   </Tab>
   <Tab title="Windows">
-    Local LLM post-processing is available on **Windows x64 and ARM64**. GPU acceleration uses NVIDIA CUDA when a compatible GPU is detected on x64; ARM64 and non-CUDA configurations fall back to CPU.
+    Local LLM post-processing is available on **Windows x64 and ARM64**. GPU acceleration on x64 uses NVIDIA CUDA and discrete AMD/Intel GPUs via Vulkan; integrated GPUs and ARM64 fall back to CPU.
```
ray-amjad: **Medium · CONFIRMED · documentation correctness** — Remove the contradictory CPU-fallback note below

This updated paragraph says discrete AMD/Intel GPUs use Vulkan, but the unchanged note at line 202 still says AMD and Intel GPUs use CPU fallback in the current Windows build. An RX 6900 XT or Intel Arc user therefore gets mutually exclusive guidance on the same page and cannot tell whether Vulkan should work.

Update the note to describe only the intentionally excluded integrated/unsupported adapters (or remove it if this paragraph already covers that distinction). Keep the wording aligned with `performance-tips.mdx`.

